PR #12, opened 2016-08-24
Remove hashie

### LINKED ISSUE #11: Remove Hashie as dependency
The gemspec has `spec.add_dependency 'hashie'`. It's only used in `Phobos.configure` and in a test (where I think it can be easily replaced with `double`).

I'd like to remove this dependency.

klippx:
Try to fix #11 

Thing with Phobos.config being a hashie mash is that it is easy to use with method calls to access things instead of using hash keys. It is also easy to mutate, which is done in the code itself (even adding methods to Hashie::Mash class)

I tried an approach to keep the method calls behaviour using a Struct.

Regarding specs: It is hard to use `double` as suggested, since it will receive a lot of "unexpected" method calls (attribute assignments). I tried an approach to use a Struct based on current configuration in phobos.yml.example, but something seems missing.

- [x] Fix broken spec
- [x] Get some feedback on how to solve this any other way


lib/phobos/deep_struct.rb
```diff
@@ -0,0 +1,16 @@
+module Phobos
+  class DeepStruct
+    def self.extract(attrs)
+      keys = attrs.keys
+      Struct.new(*keys).new(*keys.map do |k|
```
ticolucci: I was expecting the "factory" method `extract` to return a DeepStruct instance. 
So I suggest using either a module, or making it a child of Struct, so that we do have a DeepStruct instance =)
Thoughts?

klippx: Check the new version I think you will be happy ❤️

lib/phobos/deep_struct.rb
```diff
@@ -0,0 +1,29 @@
...
+            v
+          end
+
+          @hash_table[k] = v
+          new_ostruct_member(k)
+        end
+      end
+    end
+
+    def to_h
```
tulios: Since `DeepStruct` is ours, can we name this to `to_hash`? What do you think?

lib/phobos/deep_struct.rb
```diff
@@ -0,0 +1,29 @@
...
+module Phobos
+  class DeepStruct < OpenStruct
+    def initialize(hash=nil)
+      @table = {}
+      @hash_table = {}
+
+      if hash
+        hash.each_pair do |k, v|
+          k = k.to_sym
+          @table[k] = if v.is_a?(Hash)
```
alde: I'd extract this if-elsif-else chunk into a method, the initialise is a bit hard to follow

martinsvalin: I'd also use `case` instead. We only switch on one value. And `case` is nice for `is_a?` kind of conditions.

lib/phobos/deep_struct.rb
```diff
@@ -0,0 +1,29 @@
...
+            self.class.new(v)
+          elsif v.is_a?(Array) && v.all? { |el| el.is_a?(Hash) }
+            v.map { |el| self.class.new(el) }
```
martinsvalin: Why would we not deep-struct hashes in a mixed array?

klippx: now we do, spec added

lib/phobos/deep_struct.rb
```diff
@@ -0,0 +1,34 @@
...
+      if hash
+        hash.each_pair do |k, v|
+          k = k.to_sym
+          @hash_table[k] = v
+          @table[k] = to_deep_struct(v)
+        end
+      end
+    end
+
+    def to_deep_struct(v)
```
ticolucci: ❤️

lib/phobos/deep_struct.rb
```diff
@@ -0,0 +1,37 @@
...
+  class DeepStruct < OpenStruct
+    # Based on
+    # https://docs.omniref.com/ruby/2.3.0/files/lib/ostruct.rb#line=88
+    def initialize(hash=nil)
+      @table = {}
+      @hash_table = {}
+
+      if hash
+        hash.each_pair do |k, v|
+          k = k.to_sym
```
martinsvalin: This DeepStruct class is only intended for config, but the name suggests it's more general, so it may find itself used on incoming data in the future. Historically, `k.to_sym` could have opened up DoS holes. Do we care about this post Ruby 2.2?

PR #15, opened 2016-08-27
Allow configuring consumer's min_bytes and max_wait_time

sergeyevstifeev:
Apart from adding [the configuration options](https://github.com/zendesk/ruby-kafka/blob/9bb557efe545ab393b2143ac9c8a96dc7539b2e2/lib/kafka/consumer.rb#L213) from ruby-kafka, this change speeds up the test run 2x by not waiting for 5 seconds every time a listener is shut down in the tests.

### DIFF
--- a/config/phobos.yml.example
+++ b/config/phobos.yml.example
@@ -67,16 +67,16 @@ listeners:
     topic: test
     # id of the group that the consumer should join
     group_id: test-1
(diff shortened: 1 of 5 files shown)


tulios:
Nice addition but I think we should keep sane defaults in the example file. I would say the same values in ruby-kafka. What do you think?

Issue #17, opened 2016-08-31
Allow enforce payload encoding


tulios:
Allow enforce payload encoding per listener, example:

```yml
listeners:
  - handler: Phobos::EchoHandler
    topic: test
    group_id: test-1
    start_from_beginning: true
    max_bytes_per_partition: 524288
    max_concurrency: 1
    force_encoding: UTF_8
```

Where __UTF_8__ is a constant name defined in `Encoding`. We can get the correct encoding with something like:

```ruby
Encoding.get_const(listener_config.force_encoding.to_sym) if listener_config.force_encoding
```

and then encode the payload before we dispatch it to the handler, example:

```ruby
payload = message.value
if listener_config.force_encoding
  encoding = Encoding.get_const(listener_config.force_encoding.to_sym) 
  payload = payload.force_encoding(encoding)
end
```

PR #18, opened 2016-08-31
Allow enforce payload encoding

tulios:
This pull request fixes issue #17

### DIFF
--- a/config/phobos.yml.example
+++ b/config/phobos.yml.example
@@ -80,3 +80,6 @@ listeners:
     min_bytes: 1
     # Maximum duration of time to wait before returning messages from the server, in seconds
     max_wait_time: 5
+    # Apply this encoding to the message payload, if blank it uses the original encoding. This property accepts values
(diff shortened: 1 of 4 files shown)


ticolucci:
I was confused about the "empty option" in yaml. But it does gives `nil` if there's no characters in front of the option. Example:

```ruby
2.3.1 :001 > require'yaml'; YAML.load 'opt: '
 => {"opt"=>nil}
2.3.1 :002 > require'yaml'; YAML.load "opt: \n"
 => {"opt"=>nil}
2.3.1 :003 > require'yaml'; YAML.load "opt: 123\n"
 => {"opt"=>123}
2.3.1 :004 > require'yaml'; YAML.load "opt: a123\n"
 => {"opt"=>"a123"}
```

sergeyevstifeev:
One caveat is that `force_encoding` is changing the value of the object, so it actually should be `force_encoding!`. That caused a bug in the testcase - the publish was receiving a payload with UTF-8 encoding already. Maybe we should modify the listener so that it dups the object before `force_encoding` to keep the `message.value` intact.

Issue #20, opened 2016-09-07
Option to disable logging to files


tulios:
One option would be to not set "file" in the configuration file, example:

```yml
logger:
  level: info
```

PR #22, opened 2016-09-15
Option to disable logging to files

tulios:
This pull request fixes issue #20

### DIFF
--- a/spec/lib/phobos_spec.rb
+++ b/spec/lib/phobos_spec.rb
@@ -28,4 +28,18 @@ RSpec.describe Phobos do
       expect(Phobos.create_exponential_backoff).to be_a(ExponentialBackoff)
     end
   end
+
+  describe '.logger' do
+    context 'with "config.logger.file" defined' do
+      it 'writes to the logger file' do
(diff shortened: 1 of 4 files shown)


spec/lib/phobos_spec.rb
```diff
@@ -28,4 +28,18 @@
       expect(Phobos.create_exponential_backoff).to be_a(ExponentialBackoff)
     end
   end
+
+  describe '.logger' do
+    context 'with "config.logger.file" defined' do
```
sergeyevstifeev: Maybe good to have a test for the case when the `file` is not defined or is set to `false`. Something like:
```ruby
file_configs = [ false, nil ]
file_config.each do |file_config|
  Phobos.config.logger.file = file_config
  expect { Phobos.configure_logger }.to_not raise_error(SomeError)
end
```

WDYT?

tulios: Ok, I agree. I can add a new test that checks if we create the file and maybe the subdirectories, ex: `/var/logs/my-consumer/phobos.yml`

PR #25, opened 2016-10-08
Configure ruby kafka logger

tulios:
This pull request fixes issue #23

### DIFF
--- a/spec/lib/phobos_spec.rb
+++ b/spec/lib/phobos_spec.rb
@@ -34,7 +34,7 @@ RSpec.describe Phobos do
       it 'configures "logger"' do
         expect(Kafka)
           .to receive(:new)
-          .with(hash_including(logger: instance_of(Logging::Logger)))
+          .with(hash_including(Phobos.config.kafka.to_hash.merge(logger: instance_of(Logging::Logger))))
 
         Phobos.create_kafka_client
       end
--- a/spec/lib/phobos_spec.rb
+++ b/spec/lib/phobos_spec.rb
@@ -11,15 +11,33 @@ RSpec.describe Phobos do
   end
 
   describe '.create_kafka_client' do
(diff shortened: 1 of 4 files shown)


spec/lib/phobos_spec.rb
```diff
@@ -11,16 +11,34 @@
...
     it 'returns a new kafka client already configured' do
-      Phobos.configure(phobos_config_path)
+      Phobos.config.logger.ruby_kafka = nil
+      Phobos.configure_logger
 
       expect(Kafka)
         .to receive(:new)
-        .with(hash_including(Phobos.config.kafka.to_hash))
+        .with(hash_including(Phobos.config.kafka.to_hash.merge(logger: nil)))
         .and_return(:kafka_client)
```
klippx: maybe it doesnt need to return `:kafka_client`, which is a bit of a no op and also make it look like the test below

spec/lib/phobos_spec.rb
```diff
@@ -11,16 +11,34 @@
...
+    describe 'when "logger.ruby_kafka" is configured' do
+      before do
+        Phobos.config.logger.ruby_kafka = Phobos::DeepStruct.new(level: 'info')
+        Phobos.configure_logger
+      end
+
+      it 'configures "logger"' do
+        expect(Kafka)
+          .to receive(:new)
+          .with(hash_including(logger: instance_of(Logging::Logger)))
```
klippx: Does it work if you write it like the test above with a syntax as `.with(hash_including(Phobos.config.kafka.to_hash.merge(logger: instance_of(Logging::Logger))))`  ?

spec/lib/phobos_spec.rb
tulios: It's just to ensure that the method returns the kafka client

klippx: OK then, cool!

PR #30, opened 2017-08-18
Add support for hash provided settings

304:
Hello! I would like to propose an option to import settings to Phobos from a Hash-like object.

Why?
I've already have a settings managing mechanism in place (https://github.com/railsconfig/config). 
It takes care about settings inheritance and supports several environments. 
I would like to reuse my current configuration setup without adding extra `phobos.yml` files for every environment.

This PR adds an option to provide a hash to `Phobos.configure` method.

### DIFF
--- a/README.md
+++ b/README.md
@@ -248,10 +248,14 @@ MyProducer
 
 When running as a standalone service, Phobos sets up a `Listener` and `Executor` for you. When you use Phobos as a library in your own project, you need to set these components up yourself.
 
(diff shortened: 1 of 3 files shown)


tulios:
Hi @304, great!
Can you also add a bit of documentation in the section "Phobos as a library in an existing project"?

304:
@tulios, sure. I've added an usage example to README.

PR #32, opened 2017-09-21
Add support for configuring a topic prefix and group id prefix.

undergroundwebdesigns:
This functionality is useful for those using the "basic" kafka plans on Heroku,
where all topics and all consumer groups _must_ be prefixed with a given value.
See https://devcenter.heroku.com/articles/multi-tenant-kafka-on-heroku#differences-to-dedicated-kafka-plans for more.

This may also be useful on other SASS Kafka services that operate in a similar manner, or when multiple apps share the same Kafka cluster and it's desirable to prefix topics and/or consumer groups per app.

NOTE: If there's a better way to implement this (maybe the listener isn't the best place?) I'd be happy to discuss and make changes.

### DIFF
--- a/lib/phobos/listener.rb
+++ b/lib/phobos/listener.rb
@@ -10,8 +10,8 @@ module Phobos
     def initialize(handler:, group_id:, topic:, min_bytes: nil, max_wait_time: nil, force_encoding: nil, start_from_beginning: true, max_bytes_per_partition: DEFAULT_MAX_BYTES_PER_PARTITION)
       @id = SecureRandom.hex[0...6]
       @handler_class = handler
-      @group_id = group_id
-      @topic = topic
+      @group_id = prefix_group_id(group_id)
+      @topic = prefix_topic(topic)
       @subscribe_opts = {
         start_from_beginning: start_from_beginning,
         max_bytes_per_partition: max_bytes_per_partition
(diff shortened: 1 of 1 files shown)


klippx:
I think you can achieve this by using erb templating in your phobos.yml 

Just add your ENV[KAFKA_PREFIX] there and we don't need to change any code

undergroundwebdesigns:
@klippx yes you could, but with a lot of listeners that gets a little tedious, and makes the config file harder to read. For Heroku specifically where _every_ group id and topic needs to be prefixed with the same value I think having a global config for those prefixes is a nice feature to support.

klippx:
The problem of not being able to tweak prefixes for group_id and topic_id is valid, but the suggested solution is only solving this particular use case. 

The ERB templating would solve the problem but has the drawback of being tedious and harder to read. But adding code for special cases makes the Phobos code harder to read. I like config to be run once and then "frozen", it is hard to test and debug code that mutates the configuration.

Here is another though, how about to borrow an idea from Rspec configuration:

```ruby
Phobos.configure(config_path) do |config|
  config.listeners.each do |listener|
    listener.topic = prefix_topic(listener.topic)
    listener.group_id = prefix_group_id(listener.group_id)
  end
end
```

If Phobos.configure (in an optional manner) yields the parsed configuration, the user has a hook to be able to do whatever mutation on the config that they need.

Imho it would be best practices to alway read config from a file. Mutating the config should be done with caution, but at least this is done in a single place in a testable manner. It is up the the user to write the tests for this and we don't need the prefix code in Phobos.

Please let us know what you think about this idea.

klippx:
@undergroundwebdesigns Would you like to provide additional feedback to our suggestion?

undergroundwebdesigns:
@klippx that suggestion sounds like a good balance to me.

That said, I'm also fine with just closing this PR, we wound up using a different Kafka library, so it doesn't make any difference to me.

PR #35, opened 2017-10-19
Allow configuration of backoff per listener

jasondoc3:
It would be useful to have a customizable backoff per listener if desired. WDYT?

### DIFF
--- a/config/phobos.yml.example
+++ b/config/phobos.yml.example
@@ -95,3 +95,7 @@ listeners:
     # Apply this encoding to the message payload, if blank it uses the original encoding. This property accepts values
     # defined by the ruby Encoding class (https://ruby-doc.org/core-2.3.0/Encoding.html). Ex: UTF_8, ASCII_8BIT, etc
     force_encoding:
+    # Use this if custom backoff is required for a listener
+    backoff:
+      min_ms: 500
+      max_ms: 10000
--- a/lib/phobos/executor.rb
+++ b/lib/phobos/executor.rb
@@ -1,7 +1,17 @@
 module Phobos
   class Executor
     include Phobos::Instrumentation
-    LISTENER_OPTS = %i(handler group_id topic min_bytes max_wait_time force_encoding start_from_beginning max_bytes_per_partition).freeze
+    LISTENER_OPTS = %i(
+      handler
+      group_id
+      topic
+      min_bytes
+      max_wait_time
+      force_encoding
+      start_from_beginning
+      max_bytes_per_partition
+      backoff
+    ).freeze
 
     def initialize
       @threads = Concurrent::Array.new
@@ -49,7 +59,7 @@ module Phobos
 
     def run_listener(listener)
       retry_count = 0
-      backoff = Phobos.create_exponential_backoff
+      backoff = listener.create_exponential_backoff
 
       begin
         listener.start
--- a/spec/lib/phobos/executor_spec.rb
+++ b/spec/lib/phobos/executor_spec.rb
@@ -14,8 +14,20 @@ RSpec.describe Phobos::Executor do
   let(:handler2) { Phobos::EchoHandler.new }
   let(:listeners) do
     [
-      Phobos::DeepStruct.new(handler: TestHandler1.to_s, topic: topics.first, group_id: random_group_id, start_from_beginning: true, max_wait_time: MAX_WAIT_TIME),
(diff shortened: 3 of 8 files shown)


config/phobos.yml.example
```diff
@@ -95,3 +95,5 @@ listeners:
     # Apply this encoding to the message payload, if blank it uses the original encoding. This property accepts values
     # defined by the ruby Encoding class (https://ruby-doc.org/core-2.3.0/Encoding.html). Ex: UTF_8, ASCII_8BIT, etc
     force_encoding:
+    # Use this if custom backoff is required for a listener
+    backoff:
```
klippx: I think you should add 
```
  min_ms: 500
  max_ms: 30000
```
Or some other values that overrides the default, to seal the deal with this example

lib/phobos/executor.rb
```diff
@@ -1,7 +1,17 @@
...
+    LISTENER_OPTS = %i(
+      handler
+      group_id
+      topic
+      min_bytes
+      max_wait_time
+      force_encoding
+      start_from_beginning
+      max_bytes_per_partition
+      backoff
```
klippx: Down at https://github.com/klarna/phobos/pull/35/files#diff-fd46456a68de2cf7020fc43770bec636R62 in `run_listener(listener)` is where we actually use the exponential backoff we create. And right now, we still use the phobos config base defaults. I think we need to use the config for that listener (if it has custom back off configured).

maybe something like
```ruby
backoff = listener.create_exponential_backoff
```

Otherwise it will sleep for the wrong duration than the user expected.

❤️  if you manage to a test that covers this in https://github.com/klarna/phobos/blob/master/spec/lib/phobos/executor_spec.rb ?

klippx:
This PR looks really good, thank you. Very nice that you update examples, documentation and add tests. Just some small changes needed before merging, but this is a nice addition imo.

jasondoc3:
@klippx Made some changes. Let me know if I missed something or if more needs done

spec/lib/phobos/executor_spec.rb
```diff
@@ -81,4 +106,33 @@ class TestHandler2 < Phobos::EchoHandler; end
       wait_for_event('executor.stop')
     end
   end
+
+  context 'custom exponential backoff' do
```
klippx: Hmm, I am missing a negative test here (if you could call it that). We are only testing the custom. This might be a bit overwhelming but I have a few requests so bear with me, I don't think it's that much just a lot of text:

1. I would suggest to create a context backoff testing both default and custom override. Pseudo:

```ruby
context 'backoff' do
  it 'uses the default backoff settings specified in Phobos.config.backoff'
  context 'with override for a specific listener'
    it 'uses the custom backoff settings'
```

2. It would also be good to keep the concept of `listeners` (plural) and check the value of both of them, i e use `let(:listeners) { [...] }` and use at least two listeners. I think the easiest is to crash listener 1 in one test, and listener 2 in a separate test since otherwise you might fight too much with concurrency issues.

3. Please try to follow the convention of subscribing to events and waiting for them, it makes the file cleaner and consistent, and you could probably remove your `before` block which seems a bit "doing what needs to be done to make it pass" (it actually isn't really obvious what its doing or why its needed). 

4. The test is testing the wrong thing Im afraid 💀  The problem is that you are actually mocking the value - not asserting it. Never use expect().and_return - that would be asserting that rspec mocks is working. 

To verify this I tested to change the value of `min_ms: 3748` and it still passed. I tried removing the whole `backoff: { .. }` - it passed. I tried `expect_any_instance_of(ExponentialBackoff).to receive(:interval_at).once.and_return(3.0)` and... it passed. 

So... I know this isn't pretty and it is kinda hacky, but it actually asserts the correct stuff and would give reliable errors in case things breaks:

```ruby
      expect_any_instance_of(ExponentialBackoff).to receive(:interval_at) do |backoff|
        expect(backoff.instance_variable_get(:@maximum_elapsed_time)).to eq(5.678)
        expect(backoff.instance_variable_get(:@minimal_interval)).to eq(3.748)
        1
      end
```

And I use values that are a little bit more characteristic, a wild "2" or "3" might appear anywhere so I prefer more unique values. Using this way you can in fact remove the line `allow_any_instance_of(ExponentialBackoff).to receive(:randomize_factor=)` since here we are asserting before returning the value 1.

jasondoc3: @klippx Thanks for your excellent feedback and attention to detail!

I was having some trouble understanding how everything in the test file worked at first. After reviewing the file and your feedback it became apparent how the test _should_ be written. I'm also more familiar with the executor now 👍 

Third time is the charm?

klippx:
Looks great now, thanks a bunch! 💛

PR #41, opened 2017-10-25
Add hook for decoding message

klippx:
Fixes https://github.com/klarna/phobos_db_checkpoint/issues/34

- [x] Add convenient spec helper to help test consumer with minimal setup required
- [x] Add documentation for spec helper in README
- [x] Refactor listener, breaking it down to process batch and process message
- [x] Fix broken `bin/console`


CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## 1.5.0 (2017-10-25)
+
+- [enhancement] Add `decode_payload` callback to support single point of decoding a message. Useful for Handlers of messages in avro format.
```
sergeyevstifeev: It's not called `decode_payload` anymore, right?

lib/phobos.rb
```diff
@@ -9,16 +9,20 @@
...
 require 'phobos/instrumentation'
 require 'phobos/errors'
 require 'phobos/listener'
+require 'phobos/actions/process_batch'
+require 'phobos/actions/process_message'
 require 'phobos/producer'
 require 'phobos/handler'
 require 'phobos/echo_handler'
 require 'phobos/executor'
+require 'phobos/spec_helper'
```
sergeyevstifeev: why? 🤔

klippx: this is a convenient spec helper to help test consumer with minimal setup required. it is not `spec/spec_helper.rb` :)

Issue #21, opened 2016-09-07
Configurable commit strategy


tulios:
Currently, we commit after the entire batch is processed, it would be nice to have an option to define this behavior. It could be an option in the listener configuration, example:

``` yml
listeners:
  - handler: Phobos::EchoHandler
    topic: test
    # id of the group that the consumer should join
    group_id: test-1
    commit_strategy: each_batch # each_message
```

we need to give more thought in the naming used.

klippx:
Sounds great, I like the naming, makes sense.

Alternatively plainly "batch" or "message".

Even more terse just have "batch_size" and default it to 1 (same as
each_message) or >1 (each_batch with batch size defined). Assuming that batch
size can be configured.

onsdag 7 september 2016 skrev Túlio Ornelas <notifications@github.com>:

PR #42, opened 2017-11-03
Add each_message support

jasondoc3:
Addressing https://github.com/klarna/phobos/issues/21

I found that in production that I need to process using `each_message` for certain listeners. This PR adds that ability via the `consume_in_batches` listener option. I had to move some code around to accomplish this, but I think everything looks good.

A couple
1.) I added specs, but let me know if they are sufficient enough.
2.) I'm open to any configuration / naming changes

### DIFF
--- a/config/phobos.yml.example
+++ b/config/phobos.yml.example
@@ -98,8 +98,8 @@ listeners:
     # Apply this encoding to the message payload, if blank it uses the original encoding. This property accepts values
     # defined by the ruby Encoding class (https://ruby-doc.org/core-2.3.0/Encoding.html). Ex: UTF_8, ASCII_8BIT, etc
     force_encoding:
-    # Change this to false if consuming messages individually
-    consume_in_batches: true
+    # Change this to 'message' if consuming messages individually. Defaults to 'batch'
+    delivery: batch
     # Use this if custom backoff is required for a listener
     backoff:
       min_ms: 500
--- a/config/phobos.yml.example
+++ b/config/phobos.yml.example
@@ -98,7 +98,15 @@ listeners:
     # Apply this encoding to the message payload, if blank it uses the original encoding. This property accepts values
     # defined by the ruby Encoding class (https://ruby-doc.org/core-2.3.0/Encoding.html). Ex: UTF_8, ASCII_8BIT, etc
     force_encoding:
-    # Change this to 'message' if consuming messages individually. Defaults to 'batch'
+    # Specify the delivery method for a listener.
+    # Possible values: [`message`, `batch` (default)]
+    #  - `message` will yield individual messages from Ruby Kafka using `each_message` and will commit/heartbeat at every consumed message.
+    #     This is overall a bit slower than using batch, but easier to configure.
+    #  - `batch` will yield batches from Ruby Kafka using `each_batch`, and commit/heartbeat at every consumed batch.
+    #     Due to implementation in Ruby Kafka, it should be noted that when configuring large batch sizes in combination with taking long time to consume messages,
+    #     your consumers might get kicked from the Kafka cluster for not sending a heartbeat within the expected interval.
+    #     Take this into consideration when determining your configuration.
+    # Note: Ultimately commit/heartbeart will depend on the offset commit options and the heartbeat interval.
     delivery: batch
     # Use this if custom backoff is required for a listener
     backoff:
--- a/config/phobos.yml.example
+++ b/config/phobos.yml.example
@@ -98,6 +98,8 @@ listeners:
     # Apply this encoding to the message payload, if blank it uses the original encoding. This property accepts values
     # defined by the ruby Encoding class (https://ruby-doc.org/core-2.3.0/Encoding.html). Ex: UTF_8, ASCII_8BIT, etc
     force_encoding:
+    # Change this to false if consuming messages individually
+    consume_in_batches: true
     # Use this if custom backoff is required for a listener
     backoff:
       min_ms: 500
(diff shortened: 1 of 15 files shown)


tulios:
Amazing! 🎉

klippx:
This is really good stuff, you have my :+1: 

Due to new rules in protecting master this branch need to be updated, however master is currently red due to a recent merge for some reason we haven't investigated yet. We need to fix master and then this branch can be updated and merged.

jasondoc3:
@klippx Thanks! I'll be looking for master to go 💚

jasondoc3:
@tulios @klippx I actually think we should consider making `each_message` the default behavior for a listener. I know this might alter the behavior of some listeners upon upgrade, but it might be worth it.

config/phobos.yml.example
```diff
@@ -95,6 +95,8 @@ listeners:
     # Apply this encoding to the message payload, if blank it uses the original encoding. This property accepts values
     # defined by the ruby Encoding class (https://ruby-doc.org/core-2.3.0/Encoding.html). Ex: UTF_8, ASCII_8BIT, etc
     force_encoding:
+    # Change this to false if consuming messages indvidually
```
klippx: typo `indvidually`

klippx:
@jasondoc3 One benefit of introducing this change is that the default behaviour is intact. No Phobos project has ever used `each_message` and introducing a change in default behaviour could be perceived as a little bit "sneaky". I would personally prefer to keep each_batch as default at this point in time. 

Can you give some motivation why the default should be changed? We need to have a strong reason to go for this. If we decide to do this, we could introduce this change gradually with deprecation messages. 

We also need to think about what impact would this have to other configuration options such as `max_bytes_per_partition`, `min_bytes` and `max_wait_time` etc. which I think are a bit batch oriented at the moment. Maybe the README needs updates too.

1. How about changing the configuration option to be called `delivery` instead, with possible values  `[batch, message]`.
1. In `Phobos.configure`, we could detect if `delivery` is not configured at all, and if so, outputting a warning saying that the "delivery option should be specified, defaulting to batch - specify this option to silence this message".
1. If we all agree to introduce a change in the default behaviour of Phobos (we need to discuss this more) then we can also add a warning flavor to the message such as "default behavior of phobos will change to message in the future, please opt in or opt out now to silence this message".

jasondoc3:
@klippx Yeah I agree for this pull request we should keep the default behavior to prevent anything sneaky. I am bringing it up to start a discussion about potentially changing it.

**Motivation:**
When consuming in batches, consumer offsets are not committed and heartbeats are not sent to Kafka until the entire batch of messages is processed. This can be problematic if a consumer group is significantly lagging behind the head of the partition or a new consumer group is consuming from the beginning of a partition. In this case the consumer could potentially processing a huge batch of messages depending on the size of each message and the `max_bytes_per_partition` setting. If each message takes on average 0.3 seconds to process and our batch size is 1000 messages, it would take 300 seconds to process the entire batch. During this time no offsets are committed and no heartbeats are sent to kafka. This means that unless the session timeout is increased, Kafka will kick the consumers out of the group before they are finished processing the batch. When the consumer is done, it fails to commit offsets because Kafka is unaware of it. This results in a number of problems (mainly repeatedly consuming the same batch of messages) that if you aren’t aware of, can be hard to diagnose.

I believe `ruby-kafka` uses the same `fetch_batches` method under the hood for both `each_message` and `each_batch`. The difference is when it decides to commit consumer offsets and send heartbeats.

1.) I’m open to changing the configuration option.  Can do and will do.
2.) Good idea
3.) We can discuss this further if you’d like

lib/phobos/actions/process_message.rb
```diff
@@ -1,25 +1,72 @@
...
 
       def force_encoding(value)
-        @encoding ? value.force_encoding(@encoding) : value
+        @listener.encoding ? value.force_encoding(@listener.encoding) : value
+      end
+
+      def process_message(payload)
+        instrument('listener.process_message', @metadata) do |metadata|
+          time_elapsed = measure do
+            decoded_payload = @listener.handler_class.new.before_consume(payload)
```
klippx: According to issue #47 it would be great if we could instantiate a single `@listener.handler_class.new` instance and use that both for `consume` and `before_consume`.

I would also suggest to take this opportunity to change the name of `decoded_payload` to something else, maybe `massaged_payload`, `preprocessed_payload` or simply `payload` (and renaming the input payload to `input_payload`). The reasoning is that decoding a binary is only one possible use case and naming it the current way may set expectations incorrectly.

config/phobos.yml.example
```diff
@@ -95,6 +95,8 @@ listeners:
     # Apply this encoding to the message payload, if blank it uses the original encoding. This property accepts values
     # defined by the ruby Encoding class (https://ruby-doc.org/core-2.3.0/Encoding.html). Ex: UTF_8, ASCII_8BIT, etc
     force_encoding:
+    # Change this to 'message' if consuming messages individually. Defaults to 'batch'
```
klippx: Perhaps we should flesh out the documentation with everything we have discussed and learned:

```yaml
# delivery
# Specify the delivery method for a listener.
# Possible values: [`message`, `batch` (default)]
#  - `message` will yield individual messages from Ruby Kafka using `each_message` and commit/heartbeat at every consumed message. This is overall a bit slower than using batch, but easier to configure.
#  - `batch` will yield batches from Ruby Kafka using `each_batch`, and commit/heartbeat at every consumed batch. Due to implementation in Ruby Kafka, it should be noted that when configuring large batch sizes in combination with taking long time to consume messages, your consumers might get kicked out from the Kafka cluster for not sending a heartbeat within the expected interval. Take this into consideration when determining your configuration.
```

klippx:
@jasondoc3 In addition to comments, your branch also needs to be merged or rebased on master

jasondoc3:
@klippx I believe I've addressed everything. Let me know if everything looks good!

klippx:
Great PR, thank you very very much!

PR #45, opened 2017-11-09
Load boot file before configuration file

jasondoc3:
I think it may be nice to load the boot file before configuring phobos. This way any code loaded inside of the boot file can be referenced via ERB inside of `config.yml`. WDYT?

### DIFF
--- a/lib/phobos/cli/start.rb
+++ b/lib/phobos/cli/start.rb
@@ -15,12 +15,12 @@ module Phobos
       end
 
       def execute
+        load_boot_file
+
         if config_file
           validate_config_file!
           Phobos.configure(config_file)
         end
-
-        load_boot_file
 
         if listeners_file
           Phobos.add_listeners(listeners_file)


klippx:
I am entirely sure about this change, we are appending to phobos_boot for https://github.com/klarna/phobos_db_checkpoint when bootstrapping, and those lines of configuring PhobosDBCheckpoint needs Phobos already configured.

Can you explore if there is another way to achieve this without changing this order? Maybe if you require what is needed in `phobos/cli/runner`...

jasondoc3:
@klippx I'll see if I can find another way to achieve this. It's not a huge pain point for me anyways at the moment.

klippx:
Actually, after playing around with phobos CLI manually, I have found out that it is really annoying that Phobos is configured before the user has been given a chance to load his code. It ends up that you need to configure phobos twice (e.g. inside boot.rb) which is definately not a good strategy.

I think you should reopen this PR 👍🏼

PR #46, opened 2017-11-13
Add support for formatting log for stdout as json

klippx:
When running a phobos project inside a docker container, it would be nice to capture stdout as json, so splunk has a good time parsing the phobos logs.

### DIFF
--- a/config/phobos.yml.example
+++ b/config/phobos.yml.example
@@ -1,8 +1,9 @@
 logger:
   # Optional log file, set to false or remove to disable it
   file: log/phobos.log
-  # Optional log format for stdout, remove to use human readable format
-  stdout_format: json
+  # Optional output format for stdout, default is false (human readable).
+  # Set to true to enable json output.
+  stdout_json: false
   level: info
   # Comment the block to disable ruby-kafka logs
   ruby_kafka:
--- a/config/phobos.yml.example
+++ b/config/phobos.yml.example
@@ -1,6 +1,8 @@
 logger:
   # Optional log file, set to false or remove to disable it
   file: log/phobos.log
+  # Optional log format for stdout, remove to use human readable format
+  stdout_format: json
   level: info
   # Comment the block to disable ruby-kafka logs
   ruby_kafka:
--- a/lib/phobos.rb
+++ b/lib/phobos.rb
@@ -62,7 +62,7 @@ module Phobos
       date_pattern = '%Y-%m-%dT%H:%M:%S:%L%zZ'
       json_layout = Logging.layouts.json(date_pattern: date_pattern)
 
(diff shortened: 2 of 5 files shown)


config/phobos.yml.example
```diff
@@ -1,6 +1,8 @@
 logger:
   # Optional log file, set to false or remove to disable it
   file: log/phobos.log
+  # Optional log format for stdout, remove to use human readable format
+  stdout_format: json
```
martinsvalin: Is there an explicit value for human readable format? That is, if I don't want to remove the configuration but rather explicitly set it to the default.

lib/phobos.rb
```diff
@@ -60,16 +60,22 @@ def configure_logger
...
+
       appenders = [Logging.appenders.stdout(layout: stdout_layout)]
 
       Logging.backtrace(true)
       Logging.logger.root.level = silence_log ? :fatal : config.logger.level
 
       if log_file
         FileUtils.mkdir_p(File.dirname(log_file))
-        appenders << Logging.appenders.file(log_file, layout: file_layout)
+        appenders << Logging.appenders.file(log_file, layout: json_layout)
```
martinsvalin: This will change log-format in files even if you don't configure stdout format to json. The PR description doesn't mention changing the file log format.

config/phobos.yml.example
klippx: Not really. Omitting the value (i e just updating phobos and not adding any line) will keep the old behaviour. But I agree it is weird to have what looks like an Enum but really only allows one value (`json`).

Perhaps a better nomenclature would be 
```yml
logger:
  # Optional log file, set to false or remove to disable it
  file: log/phobos.log
  # Optional output format for stdout, default is false (human readable). Set to true to enable json output.
  stdout_json: false
```

lib/phobos.rb
klippx: This is just a name change to the variable. It used to be called file_layout but since the ACTUAL layout is always JSON, we think a better name is json_layout. 

In case of the name choice `stdout_layout`, this is not a fixed layout but either one of two possibilities. Hence we just called it stdout_layout to not lull anyone into thinking it is fixed.

martinsvalin: I see now.

martinsvalin:
LGTM

config/phobos.yml.example
```diff
@@ -1,6 +1,9 @@
 logger:
   # Optional log file, set to false or remove to disable it
   file: log/phobos.log
+  # Optional output format for stdout, default is false (human readable).
```
sergeyevstifeev: I think this could be rephrased to make it easier to read. I think right now it's quite hard to understand. Maybe something like this:
```Optional set to true to enable json output to stdout, default is false (human readable).```
WDYT?

Issue #48, opened 2017-11-13
Test are failing with ruby-kafka 0.5.0


TommyGustafsson:
The release of ruby-kafka 0.5.0 is not working with Phobos. Listener tests are breaking with

```shell
 IOError:
  stream closed
./vendor/bundle/ruby/2.4.0/gems/ruby-kafka-0.5.0/lib/kafka/socket_with_timeout.rb:79:in `write'
./vendor/bundle/ruby/2.4.0/gems/ruby-kafka-0.5.0/lib/kafka/socket_with_timeout.rb:79:in `write'
./vendor/bundle/ruby/2.4.0/gems/ruby-kafka-0.5.0/lib/kafka/protocol/encoder.rb:23:in `write'
./vendor/bundle/ruby/2.4.0/gems/ruby-kafka-0.5.0/lib/kafka/protocol/encoder.rb:49:in `write_int32'
```

klippx:
@jasondoc3 Continuing the discussion about pinning: I think pinning is not a solution, but a workaround. I think it is important for us to only ship Phobos with releases of RubyKafka we know are working, and since we know (by chance) that the test suite is now broken, we don't want Phobos users that run `bundle update` to update all gems to get their Phobos projects broken.

After performing this "hotfix" we need to find what the real issue is.

jasondoc3:
@klippx Thanks for the explanation. Once I saw this issue I figured that was the case 👍

PR #49, opened 2017-11-13
Add Travis integration

klippx:
- Run tests in docker (and locally)
- Remove comments from spec helper
- Pin ruby kafka to be lower than v0.5.0.
- Do not send Gemfile.lock to docker container,
as we may get a different version of installed gems compared to CI


jasondoc3:
@klippx Sorry to be a bother, but what's the reason for pinning the gem version?

jasondoc3:
Ah nvm I see the issue here. https://github.com/klarna/phobos/issues/48

PR #58, opened 2017-11-29
Change activesupport dependency from 4 to 3

mehdimu:
Hi,

I was able to run this locally with ruby 2.3.4 and Rails 3.0.5.

Had to modify the CLI to use `method_option`, because `option` wasn't working.

Let me know if you need anything else.

Thanks


klippx:
Since `option` is an alias for `method_option` I think this should be fine.

PR #64, opened 2018-01-04
Force encoding a null Kafka message.value

### LINKED ISSUE #63: Possible NoMethodError #force_encoding.
A Kafka `message.value` can be `nil`. If we set `force_encoding` on a listener, we will get a NoMethodError on nil from this line:

https://github.com/klarna/phobos/blob/39c52b56d14d26b7c1c276030a397200d85dbdad/lib/phobos/actions/process_message.rb#L54

I can PR a change to simply `value&.force_encoding(…)`, but then we'll propagate the `nil`. Is this what we want? Is there a better fix?

martinsvalin:
Fixes #63, an interaction between the `force_encoding` option and empty Kafka messages.

### DIFF
--- a/lib/phobos/actions/process_message.rb
+++ b/lib/phobos/actions/process_message.rb
@@ -51,7 +51,7 @@ module Phobos
       private
 
       def force_encoding(value)
-        @listener.encoding ? value.force_encoding(@listener.encoding) : value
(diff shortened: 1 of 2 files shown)


klippx:
Could you please also update changelog?

```
## UNRELEASED
### Fixed
- Possible NoMethodError #force_encoding #63
```

PR #71, opened 2018-02-21
Phobos overwrites ENV['RAILS_ENV'] with incorrect value.

RobCherry:
Do not set `ENV['RAILS_ENV']` or `ENV['RACK_ENV']` (see [lib/phobos.rb#L34](https://github.com/klarna/phobos/blob/v1.7.0/lib/phobos.rb#L34)).  It appears this is only used for logging purposes, and if `ENV['RAILS_ENV']` is set but `ENV['RACK_ENV']` is unset it results in `ENV['RAILS_ENV']` being set to `'development'`.  Neither of these should be set by this library.

### DIFF
--- a/lib/phobos.rb
+++ b/lib/phobos.rb
@@ -31,13 +31,12 @@ module Phobos
     attr_accessor :silence_log
 
     def configure(configuration)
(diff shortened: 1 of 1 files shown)


klippx:
Thanks for the contribution! 🐛

PR #77, opened 2018-05-15
WIP: Add headers capability to producer

klippx:
Pending merge of
- https://github.com/zendesk/ruby-kafka/pull/548

Related
-

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,7 +179,7 @@ Finally, it is also possible to preprocess the message payload before consuming 
 class MyHandler
   include Phobos::Handler
 
-  def before_consume(payload)
+  def before_consume(payload, metadata)
     # optionally preprocess payload
     payload
   end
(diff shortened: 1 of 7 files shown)


jlammers-at-ibm:
@klippx - I see in this merge the capability to access metadata / headers during consume and produce, but seems the ability so create additional metadata / headers during produce never made it to master.  Is this an oversight or is there some reason to not support metadata / headers when producing a message?

klippx:
This branch was conceived during a Hack Day about checking readiness for kafka 0.11 (ruby kafka is still not quite there)

A couple of reason why this has not been worked on:
- Lack of interest (you are the first expressing interest - welcome! 👋🏼)
- The referenced ticket was a WIP, the solution changed a bit (I think)
- Personally - in our code base/use case, our producers put the metadata in the (avro encoded) payload following a specific schema. There is nothing in metadata that we can use.

So in the end nothing was merged. I think if someone want to add it, ideally it should support both producing and consuming 👍🏼

jlammers-at-ibm:
Ok - thanks for the prompt feedback.  We are also debating putting things in the payload.

It seems the consumer code to access the metadata / headers was included in master, which is why I was so curious as to why the producer code was not.

We will discuss internally and submit a PR if we decide to pursue it - thanks...

PR #78, opened 2018-05-31
Reduce the volume of info-level log messages

ziru:
Per message/batch logging at info level can be too overwhelming.  Downgrade to debug level, to separate them from other valuable info level log messages for configure/start/stop, etc.

### DIFF
--- a/lib/phobos/listener.rb
+++ b/lib/phobos/listener.rb
@@ -103,7 +103,7 @@ module Phobos
         )
 
         batch_processor.execute
-        Phobos.logger.info { Hash(message: 'Committed offset').merge(batch_processor.metadata) }
+        Phobos.logger.debug { Hash(message: 'Committed offset').merge(batch_processor.metadata) }
         return if should_stop?
       end
     end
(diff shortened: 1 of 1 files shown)


ziru:
@klippx Are you ok with this PR?

TommyGustafsson:
I would argue against this commit since once the consumer is working, this is the only messages you will get in the logs. So it is good to have to in order to know that:

1) Your consumer is actually working and consuming messages
2) You get information on batch size,  offsets and lag which should be part of the normal logging imo

ziru:
@TommyGustafsson Include this in `config/phobos.yml` should get you all those messages.
```
logger:
  level: debug
```

In our setup, phobos emit tens of millions of log messages within a few minutes, which is the majority of the log volume.  We have to set the logger level threshold to warn.
However, using warn level threshold, we lose some important informative messages such as start/stop, etc.  Therefore, this PR use different levels, and leave it up to the phobos lib user to configure based on the need.

klippx:
Released in 1.8.0

PR #79, opened 2018-06-04
Send heartbeat while retrying on processing errors

ziru:
Also, break down big retry backoff into smaller sleep intervals, in order to send heartbeat in time and faster detection of listerner stop.

### DIFF
--- a/lib/phobos/actions/process_message.rb
+++ b/lib/phobos/actions/process_message.rb
@@ -38,10 +38,8 @@ module Phobos
               { message: "error processing message, waiting #{interval}s" }.merge(error).merge(@metadata)
             end
 
-            sleep interval
+            wait_for_next_retry interval
           end
-
-          raise Phobos::AbortError if @listener.should_stop?
 
           @metadata.merge!(retry_count: retry_count + 1)
           retry
@@ -63,6 +61,21 @@ module Phobos
             handler.consume(preprocessed_payload, @metadata)
           end
         end
+      end
+
+      def wait_for_next_retry(interval)
+        MAX_SLEEP_INTERVAL = 3
+        raise Phobos::AbortError if @listener.should_stop?
+        @listener.send_heartbeat_if_necessary
+        while interval > MAX_SLEEP_INTERVAL
+          sleep MAX_SLEEP_INTERVAL
+          raise Phobos::AbortError if @listener.should_stop?
+          @listener.send_heartbeat_if_necessary
+          interval = interval - MAX_SLEEP_INTERVAL
+        end
+        sleep interval
+        raise Phobos::AbortError if @listener.should_stop?
+        @listener.send_heartbeat_if_necessary
       end
     end
   end
--- a/lib/phobos/listener.rb
+++ b/lib/phobos/listener.rb
@@ -139,6 +139,11 @@ module Phobos
       @signal_to_stop == true
     end
 
+    def send_heartbeat_if_necessary
+      return if should_stop?
+      @consumer&.try(:send_heartbeat_if_necessary)
+    end
+
     private
 
     def listener_metadata


lib/phobos/listener.rb
```diff
@@ -139,6 +139,11 @@ def should_stop?
       @signal_to_stop == true
     end
 
+    def send_heartbeat_if_necessary
+      return if should_stop?
+      @consumer&.try(:send_heartbeat_if_necessary)
```
ziru: Related: https://github.com/zendesk/ruby-kafka/issues/355

ziru:
@klippx This PR attempts to send heartbeat more aggressively when possible.  This can help avoid rebalancing when retry backoff becomes big by informing the Kafka side we are still working on the message.

lib/phobos/actions/process_message.rb
```diff
@@ -64,6 +62,21 @@ def process_message(payload)
           end
         end
       end
+
+      def wait_for_next_retry(interval)
+        MAX_SLEEP_INTERVAL = 3
```
rstratton: Constant definition inside of methods isn't allowed in Ruby.  Could just move this into the class definition instead.

lib/phobos/actions/process_message.rb
```diff
@@ -64,6 +62,21 @@ def process_message(payload)
...
+        while interval > MAX_SLEEP_INTERVAL
+          sleep MAX_SLEEP_INTERVAL
+          raise Phobos::AbortError if @listener.should_stop?
+          @listener.send_heartbeat_if_necessary
+          interval = interval - MAX_SLEEP_INTERVAL
+        end
+        sleep interval
+        raise Phobos::AbortError if @listener.should_stop?
+        @listener.send_heartbeat_if_necessary
+      end
```
rstratton: Could be simplified to:

```ruby
def wait_for_next_retry(interval)
  while interval > 0
    raise Phobos::AbortError if @listener.should_stop?
    @listener.send_heartbeat_if_necessary
    sleep [interval, MAX_SLEEP_INTERVAL].min
    interval -= MAX_SLEEP_INTERVAL
  end
  raise Phobos::AbortError if @listener.should_stop?
  @listener.send_heartbeat_if_necessary
end
```

If you're comfortable with the following change to `Listener`, the code becomes even more readable IMO:
```ruby
# In `Listener`
def send_heartbeat_if_necessary
  raise Phobos::AbortError if should_stop?
  @consumer&.send_heartbeat_if_necessary
end

# In `ProcessMessage`
def wait_for_next_retry(interval)
  while interval > 0
    @listener.send_heartbeat_if_necessary
    sleep [interval, MAX_SLEEP_INTERVAL].min
    interval -= MAX_SLEEP_INTERVAL
  end
  @listener.send_heartbeat_if_necessary
end

File: lib/phobos/actions/process_message.rb [hunk 4]
```diff
@@ -64,6 +62,21 @@ def process_message(payload)
           end
         end
       end
+
+      def wait_for_next_retry(interval)
+        MAX_SLEEP_INTERVAL = 3
+        raise Phobos::AbortError if @listener.should_stop?
```
klippx: This snippet:
```
  raise Phobos::AbortError if @listener.should_stop?
  @listener.send_heartbeat_if_necessary
```

Let us name it "raise or send heartbeat", is repeated 3 times in this block. It feels hacky. Something simpler is necessary.

Also, I do not like mutating variables that were given as input to a function, let's create a copy of it instead. 

How about

```ruby
  def snooze(interval:)
    current_interval = interval

    while current_interval > MAX_SLEEP_INTERVAL
      sleep_and_heartbeat(MAX_SLEEP_INTERVAL)
      current_interval -= MAX_SLEEP_INTERVAL
    end

    sleep_and_heartbeat(current_interval)
  end

  private

  def sleep_and_heartbeat(interval)
    return if interval <= 0

    sleep interval

    raise Phobos::AbortError if @listener.should_stop?
    @listener.send_heartbeat_if_necessary
  end
```

klippx: `MAX_SLEEP_INTERVAL` can probably be set lower than 3, maybe even as low as `0.1`. One could think it should be configurable but I think too much configuration options can be bad especially when it is hard to explain what the option actually does.

I am not sure if there are performance implications of running 30 sleeps of 0.1 instead of 1 sleep of 3, but the lines of the block we are invoking is guarded by if statements so I am not worried about the block statements as such. Might want to investigate this.

klippx:
@ziru Ping

lib/phobos/actions/process_message.rb
rstratton: Hi @klippx, I'm looking at picking up where Ziru left off on this PR, curious to hear why it would be desirable to set the `MAX_SLEEP_INTERVAL` as low as `0.1`, is that just to accommodate users of Phobos who have a very short `session_timeout` configured?

Your other comments make sense, I'll work on addressing them and submit another PR.

PR #81, opened 2018-07-06
Allow a custom logger for both Phobos and RubyKafka.

dorner:
I have a custom Rails logger I'm using which uses ActiveSupport tagged logging and not the `Logging` framework. This PR allows me to set the custom logger to Phobos instead of being required to use Logging.

### DIFF
--- a/README.md
+++ b/README.md
@@ -320,17 +320,24 @@ The configuration file is organized in 6 sections. Take a look at the example fi
 
 The file will be parsed through ERB so ERB syntax/file extension is supported beside the YML format.
 
-__logger__ configures the logger for all Phobos components, it automatically outputs to `STDOUT` and it saves the log in the configured file
+__logger__ configures the logger for all Phobos components. It automatically 
+outputs to `STDOUT` and it saves the log in the configured file.
 
-__kafka__ provides configurations for every `Kafka::Client` created over the application. All [options supported by  `ruby-kafka`][ruby-kafka-client] can be provided.
+__kafka__ provides configurations for every `Kafka::Client` created over the application. 
+All [options supported by  `ruby-kafka`][ruby-kafka-client] can be provided.
 
-__producer__ provides configurations for all producers created over the application, the options are the same for regular and async producers. All [options supported by  `ruby-kafka`][ruby-kafka-producer] can be provided.
+__producer__ provides configurations for all producers created over the application, 
+the options are the same for regular and async producers. 
+All [options supported by  `ruby-kafka`][ruby-kafka-producer] can be provided.
 
-__consumer__ provides configurations for all consumer groups created over the application. All [options supported by  `ruby-kafka`][ruby-kafka-consumer] can be provided.
+__consumer__ provides configurations for all consumer groups created over the application. 
+All [options supported by  `ruby-kafka`][ruby-kafka-consumer] can be provided.
 
-__backoff__ Phobos provides automatic retries for your handlers, if an exception is raised the listener will retry following the back off configured here. Backoff can also be configured per listener.
+__backoff__ Phobos provides automatic retries for your handlers. If an exception 
+is raised, the listener will retry following the back off configured here. 
+Backoff can also be configured per listener.
 
-__listeners__ is the list of listeners configured, each listener represents a consumers group
+__listeners__ is the list of listeners configured. Each listener represents a consumer group.
 
 [ruby-kafka-client]: http://www.rubydoc.info/gems/ruby-kafka/Kafka%2FClient%3Ainitialize
 [ruby-kafka-consumer]: http://www.rubydoc.info/gems/ruby-kafka/Kafka%2FClient%3Aconsumer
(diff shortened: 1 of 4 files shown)


lib/phobos.rb
```diff
@@ -77,16 +78,20 @@ def configure_logger
         appenders << Logging.appenders.file(log_file, layout: json_layout)
       end
 
-      @ruby_kafka_logger = nil
+      @ruby_kafka_logger = config.custom_kafka_logger
 
-      if ruby_kafka
+      if ruby_kafka && @ruby_kafka_logger.nil?
```
klippx: to be in line with the newly added block below, perhaps do

```
if config.custom_kafka_logger
  @ruby_kafka_logger = config.custom_kafka_logger
else
  ...
end
```

lib/phobos.rb
```diff
@@ -77,16 +78,20 @@ def configure_logger
         appenders << Logging.appenders.file(log_file, layout: json_layout)
       end
 
-      @ruby_kafka_logger = nil
+      @ruby_kafka_logger = config.custom_kafka_logger
 
-      if ruby_kafka
```
klippx: Do you think it is reasonable to add some additional sections to README?

klippx: Do you think it would be a good idea to describe (and protect) your changes with some specs?

klippx: If you would like to update the changelog, that would be appreciated

klippx: Finally, CodeClimate reported some complaints, is this something you would be up to have a look at?

klippx:
The comments are general comments, not comment about the line `if ruby_kafka` 🙄

lib/phobos.rb
dorner: sure!

lib/phobos.rb
dorner: absolutely - didn't occur to me before :)

dorner: Yep, I can do that. Wanted to put up the PR for an initial look first.

dorner: sure, I'll do all the housekeeping around this PR if everything else looks good. 👍

dorner:
@klippx I've added a couple of extra commits. Can you take a look?

lib/phobos.rb
```diff
@@ -55,38 +55,51 @@ def create_exponential_backoff(backoff_config = nil)
...
+      elsif ruby_kafka
         @ruby_kafka_logger = Logging.logger['RubyKafka']
         @ruby_kafka_logger.appenders = appenders
         @ruby_kafka_logger.level = silence_log ? :fatal : ruby_kafka.level
       end
 
-      @logger = Logging.logger[self]
-      @logger.appenders = appenders
+      custom_logger = config.custom_logger
+      if custom_logger
```
klippx: It would be cleaner to have

```ruby
if config.custom_kafka_logger
  @ruby_kafka_logger = config.custom_kafka_logger

...

if config.custom_logger
  @logger = config.custom_logger
```

lib/phobos.rb
```diff
@@ -55,38 +55,51 @@ def create_exponential_backoff(backoff_config = nil)
...
-      appenders = [Logging.appenders.stdout(layout: stdout_layout)]
 
       Logging.backtrace(true)
       Logging.logger.root.level = silence_log ? :fatal : config.logger.level
-
-      if log_file
-        FileUtils.mkdir_p(File.dirname(log_file))
-        appenders << Logging.appenders.file(log_file, layout: json_layout)
-      end
+      appenders = logger_appenders
```
klippx: I think you can remove this, and change line 73 to `@ruby_kafka_logger.appenders = logger_appenders`

lib/phobos.rb
```diff
@@ -55,38 +55,51 @@ def create_exponential_backoff(backoff_config = nil)
...
+      custom_logger = config.custom_logger
+      if custom_logger
+        @logger = custom_logger
+      else
+        @logger = Logging.logger[self]
+        @logger.appenders = appenders
+      end
+    end
+
+    def logger_appenders
```
klippx: nice!

spec/lib/phobos_spec.rb
```diff
@@ -33,6 +33,32 @@
         expect(Phobos.config.kafka.client_id).to eq('client_id')
         expect(Phobos.config.logger.file).to eq('log/phobos.log')
       end
+
+      it 'sets custom loggers' do
```
klippx: nice, much appreciated

README.md
```diff
@@ -350,6 +357,23 @@ $ phobos start -c /var/configs/my.yml -l /var/configs/additional_listeners.yml
 Note that the config file _must_ still specify a listeners section, though it
 can be empty.
 
+#### Custom configuration/logging
```
klippx: nice!

lib/phobos.rb
dorner: It gets used twice, once on line 73 and once on line 82, so it's memoized this way :)

lib/phobos.rb
dorner: Ehh, six of one, half a dozen of the other. Saves 2 lines but also adds 2 extra pointer moves. I can change it :)

dorner: Heh, changing this added more CodeClimate complaints :)

dorner:
@klippx one more commit added!

klippx:
Thank you for the contribution!

klippx:
Released in 1.8.0

dorner:
Awesome!

PR #82, opened 2018-07-11
Around consume instance method

michberr:
## Reasoning:
1) **Consistency/readability**: It's generally confusing that `#before_consume` is an instance method on the handler and `.around_consume` is a class method. 

2) **Flexibility**: Allows users to set custom instance variables on the handler that will be accessible in around_consume.

This feature has come up consistently during development. For example, we have built some custom logic into around_consume that relies on metadata from the original payload. We lose that metadata in the before_consume step, forcing us to put all of the logic into around_consume. Another example is error handling for before_consume. In some cases we want to catch exceptions in before_consume and continue to proceed with message processing. Setting an instance variable on the handler is a logical way to signal this downstream.

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,7 +198,7 @@ Finally, it is also possible to preprocess the message payload before consuming 
 class MyHandler
   include Phobos::Handler
 
-  def before_consume(payload)
+  def before_consume(payload, metadata)
     # optionally preprocess payload
     payload
   end
@@ -525,7 +525,7 @@ describe MyConsumer do
 
   it 'consumes my message' do
     expect_any_instance_of(described_class).to receive(:around_consume).with(payload, metadata).once.and_call_original
-    expect_any_instance_of(described_class).to receive(:before_consume).with(payload).once.and_call_original
+    expect_any_instance_of(described_class).to receive(:before_consume).with(payload, metadata).once.and_call_original
     expect_any_instance_of(described_class).to receive(:consume).with(payload, metadata).once.and_call_original
 
     process_message(handler: described_class, payload: payload, metadata: metadata)
--- a/README.md
+++ b/README.md
@@ -156,7 +156,25 @@ class MyHandler
 end
 ```
 
-It is also possible to control the execution of `#consume` with the class method `.around_consume(payload, metadata)`. This method receives the payload and metadata, and then invokes `#consume` method by means of a block; example:
+It is also possible to control the execution of `#consume` with the method `#around_consume(payload, metadata)`. This method receives the payload and metadata, and then invokes `#consume` method by means of a block; example:
+
+```ruby
+class MyHandler
+  include Phobos::Handler
+
+  def around_consume(payload, metadata)
+    Phobos.logger.info "consuming..."
+    output = yield
+    Phobos.logger.info "done, output: #{output}"
+  end
+
+  def consume(payload, metadata)
+    # consume or skip message
+  end
+end
+```
+
+Note: `around_consume` can also be defined as a class method. If defined, the class method `.around_consume` takes precedence.
 
 ```ruby
 class MyHandler
@@ -174,7 +192,7 @@ class MyHandler
 end
 ```
 
-Finally, it is also possible to preprocess the message payload before consuming it using the `before_consume` hook which is invoked before `.around_consume` and `#consume`. The result of this operation will be assigned to payload, so it is important to return the modified payload. This can be very useful, for example if you want a single point of decoding Avro messages and want the payload as a hash instead of a binary.
+Finally, it is also possible to preprocess the message payload before consuming it using the `before_consume` hook which is invoked before `#around_consume` and `#consume`. The result of this operation will be assigned to payload, so it is important to return the modified payload. This can be very useful, for example if you want a single point of decoding Avro messages and want the payload as a hash instead of a binary.
 
 ```ruby
 class MyHandler
@@ -195,7 +213,7 @@ The hander life cycle can be illustrated as:
 
 or optionally,
 
-  `.start` -> `#before_consume` -> `.around_consume` [ `#consume` ] -> `.stop`
+  `.start` -> `#before_consume` -> `#around_consume` [ `#consume` ] -> `.stop`
 
 ### <a name="usage-producing-messages-to-kafka"></a> Producing messages to Kafka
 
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -174,13 +192,13 @@ class MyHandler
...
 
-Finally, it is also possible to preprocess the message payload before consuming it using the `before_consume` hook which is invoked before `.around_consume` and `#consume`. The result of this operation will be assigned to payload, so it is important to return the modified payload. This can be very useful, for example if you want a single point of decoding Avro messages and want the payload as a hash instead of a binary.
+Finally, it is also possible to preprocess the message payload before consuming it using the `before_consume` hook which is invoked before `#around_consume` and `#consume`. The result of this operation will be assigned to payload, so it is important to return the modified payload. This can be very useful, for example if you want a single point of decoding Avro messages and want the payload as a hash instead of a binary.
 
 ```ruby
 class MyHandler
   include Phobos::Handler
 
-  def before_consume(payload)
+  def before_consume(payload, metadata)
```
michberr: This is a breaking change. I would like to have access to the metadata inside of before_consume for logging purposes (right now i'm digging into the Phobos.config to get that information). My preference would be to instantiate the handler with the metadata so there's no need to pass it as an arg to `before_consume`, `around_consume`, or `consume`:

```
def process_message(payload)
  instrument('listener.process_message', @metadata) do
    handler = @listener.handler_class.new(@metadata)
    consume_block = Proc.new { handler.consume(preprocessed_payload) }

    if @listener.handler_class.respond_to?(:around_consume)
      # around_consume class method implementation
      @listener.handler_class.around_consume(preprocessed_payload, &consume_block)
    else
      # around_consume instance method implementation
      handler.around_consume(preprocessed_payload, &consume_block)
    end
  end
end
```
However, this is a larger breaking change. Open other perspectives on this.

klippx:
Hi @michberr and welcome to Phobos!

Spontaneosly I like the idea, and I agree with your reasoning. I will have to look into what this means for us and especially for https://github.com/klarna/phobos_db_checkpoint 

I am on vacation 🏝 right now so response time might be varying.

klippx:
Question to you @michberr and @jasondoc3 

Given that this change is merged in the future:
Could this, from your perspective, also mean removing `#before_consume` since that can be part of `#around_consume` now...? 🤔 

I am not to fond of `#before_consume`, it is very dodgy, but as Phobos works atm we need it right now.

klippx:
I moved the discussion about before consume to a separate issue, see klarna/phobos#84

lib/phobos/actions/process_message.rb
```diff
@@ -57,10 +57,15 @@ def force_encoding(value)
...
         instrument('listener.process_message', @metadata) do
           handler = @listener.handler_class.new
-          preprocessed_payload = handler.before_consume(payload)
+          preprocessed_payload = handler.before_consume(payload, @metadata)
+          consume_block = Proc.new { handler.consume(preprocessed_payload, @metadata) }
 
-          @listener.handler_class.around_consume(preprocessed_payload, @metadata) do
-            handler.consume(preprocessed_payload, @metadata)
+          if @listener.handler_class.respond_to?(:around_consume)
+            # around_consume class method implementation
```
klippx: Maybe emit some kind of deprecation, such as

```
warn Kernel.caller.first + " around_consume has been moved to instance method, please update your consumer. This will not be backwards compatible in the future."
```

I am not sure about how to best emit deprecation warnings or what the message should say, just a suggestion.

README.md
klippx: There are 2 topics of discussion here.

### This code/The breaking change 

Well, I guess it will be a bumpy ride since we might want to remove `before_consume` in the future, but oh well, it was our bad for not adding metadata to it, in the first place.

### About the suggestion/preference
At first glance it looks clean, I like it. 

But it could be asked by someone else, why not also send payload as argument to initialize? But we happen to know that sometimes we want to change payload (decode and whatnot). So... could a third or fourth party potentially want to change/mutate metadata? 🤔 Also, in practice, will it not look a bit weird that we have methods with `metadata` appearing from nowhere, such as:

```ruby
def around_consume(payload)
  payload = decode_payload(payload)
  logger.log(payload: payload)
  logger.log(metadata: metadata)
end
```

Where did metadata come from? Looks a bit magic. Finally, since we discussed yielding payload to from around_consume, maybe we want to yield metadata as well here to not make the same mistake as we did (ignoring metadata)

Maybe you can create a new Issue for passing the `@metadata` to the initializer of the handler class and let's continue the discussion there.

michberr: Sounds good! Yes i realize this PR really opens up a whole can of worms on discrete but related design decisions. I will work on a new issue for discussing the instantiation of the handler.

lib/phobos/actions/process_message.rb
```diff
@@ -57,10 +57,17 @@ def force_encoding(value)
...
+            warn "#{Kernel.caller.first} around_consume has been moved to instance method, please update your consumer."\
```
klippx: I think we can follow the same pattern as in PhobosDBCheckpoint here, namely define a deprecate method in Phobos module

```
    def deprecate(message)
      warn "DEPRECATION WARNING: #{message} #{Kernel.caller.first}"
    end
```

Then we can call it using `Phobos.deprecate("around_consume has been moved to instance method, please update your consumer " \ .....)`

It is easier to have one single point of doing deprecations and prefixing with DEPRECATION WARNING catches one's eye more reliably.

lib/phobos/actions/process_message.rb
```diff
@@ -57,10 +57,17 @@ def force_encoding(value)
       def process_message(payload)
         instrument('listener.process_message', @metadata) do
           handler = @listener.handler_class.new
-          preprocessed_payload = handler.before_consume(payload)
+          preprocessed_payload = handler.before_consume(payload, @metadata)
```
klippx: In order to be backwards compatible here we can do 

```ruby
          preprocessed_payload = begin
            handler.before_consume(payload, @metadata)
          rescue ArgumentError => e
            Phobos.deprecate("before_consume now expects metadata as second argument, please update your consumer. This will not be backwards compatible in the future.")
            handler.before_consume(payload)
          end
```

lib/phobos/actions/process_message.rb
michberr: Done!

lib/phobos/actions/process_message.rb
michberr: Done!

michberr:
@klippx having some issues with Travis. The first build failed for some reason I can't comprehend (undefined method 'debug' for logging::logger)https://travis-ci.com/klarna/phobos/builds/79593666. I triggered a rebuild by pushing an empty commit which then passed https://travis-ci.com/klarna/phobos/builds/79595739, but github the PR is still pointing to the failed build.

klippx:
Yes, that is a separate issue which is hard to recreate. We should track it
and fix it. I think it has to do with concurrency and logging not being
thread safe. We are seeing it in production for a different project but
have not tackled it yet. Only speculation from my side...
fre 20 juli 2018 kl. 18:47 skrev Michelle Berry <notifications@github.com>:

klippx:
Thanks for the effort, really good first PR ❤️ I will update changelog.

klippx:
Released in 1.8.0

PR #83, opened 2018-07-13
Send heartbeat while retrying on processing errors

rstratton:
Also, break down big retry backoff into smaller sleep intervals, in
order to send heartbeat in time and faster detection of listerner
stop.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Changed
 - Reduce the volume of info-level log messages #78
 - Add ability to configure a custom logger #81
+- Send consumer heartbeats between retry attempts #83
 
 ## [1.7.2] - 2018-05-03
 ### Added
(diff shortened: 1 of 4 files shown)


rstratton:
This PR attempts to address the feedback in https://github.com/klarna/phobos/pull/79

rstratton:
@klippx copying my question from the other PR to consolidate our conversation:

> I'm curious to hear why it would be desirable to set the `MAX_SLEEP_INTERVAL` as low as `0.1`, is that just to accommodate users of Phobos who have a very short `session_timeout` configured?

klippx:
I like your initiative 👍🏼 

About the `0.1` - never mind - I think this is fine.

Would you mind updating the changelog as well?

PS: Are you a colleague of @ziru?

rstratton:
Thanks for the review!  I've updated the changelog, let me know if you'd prefer different wording.  And yes, I'm a colleague of @ziru, as well as @michberr and @jasondoc3 😄

klippx:
Thanks for the contrib, and welcome to Phobos! ❤️

klippx:
Released in 1.8.0

PR #86, opened 2018-07-24
Enable partition keys to be separate from payload keys

dorner:
Currently, Phobos only uses message/payload keys. It automatically sets the partition key to be the same as the payload key.

There are use cases where these should be separate. For example, say there is a comment on a post. The Comments topic would want the comment ID to be the payload key (since it indicates that all updates to that individual comment should be compacted) while the partition key would likely be the post (so all actions related to the same post are put in the same partition).

This PR adds this functionality into Phobos.

### DIFF
--- a/spec/lib/phobos/producer_spec.rb
+++ b/spec/lib/phobos/producer_spec.rb
@@ -16,7 +16,7 @@ RSpec.describe Phobos::Producer do
 
       expect(TestProducer1.producer)
         .to receive(:publish_list)
-        .with([{ topic: 'topic', payload: 'message', key: 'key' }])
+        .with([{ topic: 'topic', payload: 'message', key: 'key', partition_key: nil }])
 
       subject.producer.publish('topic', 'message', 'key')
     end
@@ -29,10 +29,10 @@ RSpec.describe Phobos::Producer do
 
       expect(TestProducer1.producer)
         .to receive(:async_publish_list)
-        .with([{ topic: 'topic', payload: 'message', key: 'key' }])
+        .with([{ topic: 'topic', payload: 'message', key: 'key', partition_key: 'partition_key' }])
 
       TestProducer1.producer.create_async_producer
-      subject.producer.async_publish('topic', 'message', 'key')
+      subject.producer.async_publish('topic', 'message', 'key', 'partition_key')
     end
   end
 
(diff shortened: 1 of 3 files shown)


spec/lib/phobos/producer_spec.rb
```diff
@@ -16,7 +16,7 @@ class TestProducer1
 
       expect(TestProducer1.producer)
         .to receive(:publish_list)
-        .with([{ topic: 'topic', payload: 'message', key: 'key' }])
+        .with([{ topic: 'topic', payload: 'message', key: 'key', partition_key: nil }])
 
       subject.producer.publish('topic', 'message', 'key')
```
klippx: shouldn't this spec be updated use the new 4th argument?

dorner: Nope, I still want to make sure that for backwards compatibility it still works correctly if no partition key is given.

klippx:
I am still on vacation 🌴 but will have a look 👀 asap!
fre 10 aug. 2018 kl. 17:21 skrev Daniel Orner <notifications@github.com>:

ziru:
I'm not sure whether this change is a good move, though the separation of key/partition_key might stem from ruby-kafka.

Reason:
- if the kafka consumer needs to publish the message to another topic (ex: pipelining/splitting), how could the downstream consumer knows the partition_key (only the key/value is passed around, and partition_key is only used by ruby-kafka to compute the partition number)?

Therefore, the `key` used for kafka publishing has special meaning for partitioning.
For pure payload keys, better include it in the value field of the message, rather than the key field.

dorner:
@ziru the main reason to separate them is because the key is used for more than partitioning. It's essentially a way to uniquely identify a record. E.g. when compacting records, the key is used to determine which records are the "same" and can be easily deleted.

The partition is different from this - it's used to group messages together that should be consumed together. E.g. you might want to process all posts related to the same user together, but you wouldn't want to blow away all posts whenever any post is updated.

ziru:
@dorner Kafka log compaction only retains latest value per key per partition.   Thus, the separation might cause values for the same key not compacted across partitions, right?

dorner:
@ziru you're correct - that would be up to the user to make sure that they pick the right partition key such that you should never have the same key associated with different partition keys. You can think of it as a key being a "unique ID" and a partition being a "group ID" - if a post always belongs to a user, you can safely partition by user because a post would never be in any partition for a different user.

Most of the time, using the key as the partition key is fine, which is why it's still the default.

Issue #98, opened 2018-12-12
Fix crash on startup with namespaced handlers


dorner:
If I have a handler called something like `Handlers::MyHandler`, Phobos 1.8.1 crashes. This was introduced with this commit: https://github.com/klarna/phobos/commit/c38d87f38e426b28e187d27b76a2ce615a03d20b

PR incoming.

dorner:
Thanks!

PR #99, opened 2018-12-12
#98: Fix crash on startup with namespaced handlers

dorner:
Issue was introduced by switching `constantize` in for `Object.const_defined?`. This only looks at the top level and doesn't try to load namespaces.

### DIFF
--- a/lib/phobos/cli/start.rb
+++ b/lib/phobos/cli/start.rb
@@ -39,7 +39,11 @@ module Phobos
         Phobos.config.listeners.each do |listener|
           handler = listener.handler
 
-          Object.const_defined?(handler) || error_exit("Handler '#{handler}' not defined")
+          begin
+            handler.constantize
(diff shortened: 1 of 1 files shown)


klippx:
Thanks, this is good. I guess rubocop had some complaints about this however;

```
       +lib/phobos/cli/start.rb:38:7: C: Metrics/MethodLength: Method has too many lines. [18/15]
       +      def validate_listeners! ...
       +      ^^^^^^^^^^^^^^^^^^^^^^^
```

But it would be OK to just drop a `# rubocop:disable Metrics/MethodLength:` before method definition and ``# rubocop:enable Metrics/MethodLength:` again after the method definition.

dorner:
@klippx fixed!

dorner:
Nice! Let me know when we're up for a version bump @klippx 👍

PR #101, opened 2018-12-26
Introduce true batch consume method

dorner:
As previously discussed, this PR provides a method for passing a full batch of messages into a handler rather than a single message at a time. This can help speed up things like database inserts (we can do a mass insert instead of inserting one message at a time).

Feedback welcome! @klippx


lib/phobos/actions/process_message_batch.rb
```diff
@@ -0,0 +1,65 @@
+# frozen_string_literal: true
+
+module Phobos
+  module Actions
+    class BatchMessage
```
klippx: Should preferably be in its own file called `batch_message.rb` placed in `lib/phobos`, it is much easier to find and write isolated tests etc for it.

lib/phobos/actions/process_message_batch.rb
```diff
@@ -0,0 +1,65 @@
...
+      def initialize(key:, partition:, offset:, payload:)
+        @key = key
+        @partition = partition
+        @offset = offset
+        @payload = payload
+      end
+
+      def ==(other)
+        [:key, :partition, :offset, :payload].all? do |s|
+          send(s) == other.send(s)
```
klippx: does it run with `public_send` over `send`? If so, that would be preferred.

lib/phobos/actions/process_message_batch.rb
```diff
@@ -0,0 +1,65 @@
...
+        end
+      end
+    end
+
+    class ProcessMessageBatch < ProcessMessage
```
klippx: You are subclassing ProcessMessage probably to get access to retry/handle error, even though conceptually they are very different. However I feel this is kind of a shortcut and it could be worth investigating for various reasons:
- you get stuff you dont want/need such as `force_encoding`, duplicating a constant `MAX_SLEEP_INTERVAL`
- the code is incrementing instrumentation regarding `listener.process_message` but this is clearly not the case, this is `listener.process_message_batch`
- the code is incrementing instrumentation regarding `listener.retry_handler_error` but in this case I think it should be `listener.retry_handler_error_batch` and the message should mention `error processing batch` instead.
- I would personally move error handling in `process_message.rb` to a module `process_error_handling.rb`, specifically the methods `retry_count`, `increment_retry_count`, `backoff`, `backoff_interval`, `snooze` looks like they can be moved easily. Only `handle_error` needs special attention with its logging and instrumentation. could possibly stay in `process_message.rb` but reimplemented in `process_message_batch.rb` using correct instrumentation and logging.
- include the ProcessErrorHandling and then remove the subclassing
- update the Readme to document the 2 new instrumentation options available

README.md
```diff
@@ -214,6 +214,38 @@ The hander life cycle can be illustrated as:
...
+    end
+  end
+  
+  def around_consume_batch(payloads, metadata)
+    yield
+  end
+  
+  def consume_batch(payloads, metadata)
+    payloads.each do |p|
+      logger.info("Got payload #{p.payload}, #{p.partition}, #{p.offset}, #{p.key}")
```
klippx: just for the kicks, add `#{p[:timestamp]}` ?

lib/phobos/listener.rb
```diff
@@ -123,7 +123,14 @@ def stop_listener
     end
 
     def start_consumer_loop
-      @delivery == 'batch' ? consume_each_batch : consume_each_message
+      # validate batch handling
+      if @delivery == 'batch'
+        consume_each_batch
+      elsif handler_class.method_defined?(:consume)
```
klippx: This is a bit complicated for me. I would be open to have a new delivery option `batch_inline` since I think this makes the readme section more clear. Example `phobos.yml`:

```
listeners:
  - handler: MyOldBatchHandler
     delivery: batch
  - handler: MyNewBatchHandler
     delivery: batch_inline
```

When opening `MyNewBatchHandler` I would expect to see `include Phobos::BatchHandler`, and when opening `MyOldBatchHandler` I would expect to see the old classic `include Phobos::Handler`.

This would make this `if` block much clearer and more [#batch_inline]

Of course, the delivery option `batch_inline` must be documented in the README but I hope you agree this will make adoption much easier for other people.

lib/phobos/actions/process_batch.rb
```diff
@@ -20,15 +20,31 @@ def initialize(listener:, batch:, listener_metadata:)
...
       def execute
         instrument('listener.process_batch', @metadata) do |_metadata|
-          @batch.messages.each do |message|
-            Phobos::Actions::ProcessMessage.new(
-              listener: @listener,
-              message: message,
-              listener_metadata: @listener_metadata
-            ).execute
+          if @listener.handler_class.method_defined?(:consume_batch)
+            process_batch_inline
```
klippx: This is also a bit error prone way of inferring the control flow (I have a related comment below mentioning `batch_inline` as an explicit delivery option [#batch_inline])

The thing this idea could enable is to initialize `ProcessBatch` with delivery method = one of `batch` / `batch_inline`. This would change the signature of initialize, but you can always default this to `batch`. And btw please make them constants in this file for now, eg `const DELIVERY_BATCH = 'batch'` and `const DELIVERY_INLINE = 'inline_batch'`

Then you can be more clear here in this if statement say with confidence:

`if @delivery_option == DELIVERY_INLINE`

spec/lib/phobos/listener_spec.rb
```diff
@@ -119,6 +123,16 @@ class TestListenerHandlerWithProducer < TestListenerHandler
     end
   end
 
+  describe 'errors with invalid handlers' do
+    let(:handler_class) { TestBatchListenerHandler }
+    let(:delivery) { 'message' }
+
+    it 'should raise an error if given batch handler in single message mode' do
+      expect { listener.start }.to raise_error(Phobos::InvalidHandlerError,
```
klippx: not sure if we need this test if we have the proposed solution `batch_inline` as delivery option [#batch_inline]

lib/phobos/actions/process_batch.rb
```diff
@@ -20,15 +20,31 @@ def initialize(listener:, batch:, listener_metadata:)
 
       def execute
         instrument('listener.process_batch', @metadata) do |_metadata|
-          @batch.messages.each do |message|
-            Phobos::Actions::ProcessMessage.new(
-              listener: @listener,
-              message: message,
-              listener_metadata: @listener_metadata
-            ).execute
+          if @listener.handler_class.method_defined?(:consume_batch)
```
klippx: This is also a bit error prone way of inferring the control flow (I have a related comment below mentioning `batch_inline` as an explicit delivery option [#batch_inline])

The thing this idea could enable is to initialize `ProcessBatch` with delivery method = one of `batch` / `batch_inline`. This would change the signature of initialize, but you can always default this to `batch`. And btw please make them constants in this file for now, eg `const DELIVERY_BATCH = 'batch'` and `const DELIVERY_INLINE = 'inline_batch'`

Then you can be more clear here in this if statement say with confidence:

`if @delivery_option == DELIVERY_INLINE`

klippx:
Another question on the top of my head is how does error handling actually work in the case of true batch consumption?

Let's say we get a batch of 100 messages, they are delivered inline to the consumer which includes `Phobos::BatchHandler`. For some reason this fails on message 50 because the user is deserializing `payloads` into `payloads.each do |payload|` thus handling them one by one. But it has done some non-idemptotent operations based on those 50 messages. 

=> I guess retry of the batch will actually go through all those 100 messages again.

If this is an example of bad practice (aka a "DO NOT") then maybe we should state that as a warning in the readme.

spec/lib/phobos/listener_spec.rb
dorner: Yep, I'll delete.

dorner:
@klippx PR is updated - please take another look!

klippx:
You created a new module (as per request, thank you) but you put it in the `actions` folder (this is about `lib/phobos/actions/processor.rb`) and this folder should only contain "actions". So the suggestion would be to move it down one step.

klippx:
I really like the PR so far, it is coming along great!

But since action folder was brought to my attention again it got me thinking about actions. Would you think it would make sense to make process_batch_inline it's own action? Then we could leave `process_batch.rb` alone, and get rid of the `@inline` parameter.

And instead put the burden of effort in `lib/phobos/listener.rb`:

```
    def start_consumer_loop
      @delivery == 'batch' ? consume_each_batch : consume_each_message
      # validate batch handling
      case @delivery
      when 'batch'
        consume_each_batch
      when 'inline_batch'
        consume_each_batch_inline
      else
        consume_each_message
      end
    end
```

and

```
def consume_each_batch_inline
      @consumer.each_batch(@message_processing_opts) do |batch|
        batch_processor = Phobos::Actions::ProcessBatchInline.new(
          listener: self,
          batch: batch,
          listener_metadata: listener_metadata
        )

        batch_processor.execute
        log_debug('Committed offset', batch_processor.metadata)
        return nil if should_stop?
      end
    end
```

klippx:
Thanks for a great job so far and for being open minded about the reviews :) This is super close to merging now

lib/phobos/actions/process_message_batch.rb
dorner: So re a separate `process_batch_inline` - I didn't really see the value in a separate instrumentation here because literally all that happens was it went into ProcessBatch and then immediately went into ProcessBatchMessages. That's also why I called it `process_messages` - it's a relic from when it lived inside ProcessBatch.

Based on your comment below where we skip ProcessBatch entirely, this all works quite nicely. :) I'll make the changes.

dorner:
Those are great ideas and make some of the weirdnesses less weird. :p I'll make those changes now.

dorner:
@klippx one more commit - I think everything should be covered now! Thanks for the tips - this is way cleaner than my original implementation 👍

lib/phobos/actions/process_batch_inline.rb
```diff
@@ -0,0 +1,63 @@
...
+        end
+      end
+
+      private
+
+      def process_batch(payloads)
+        instrument('listener.process_batch_inline', @metadata) do |_metadata|
+          handler = @listener.handler_class.new
+
+          preprocessed_payloads = before_consume(handler, payloads)
```
klippx: I think we should call `preprocessed_payloads = handler.before_consume_batch(payloads, @metadata)` directly, and skip the roundtrip to the function below (which can be deleted).

README.md
```diff
@@ -214,6 +214,46 @@ The hander life cycle can be illustrated as:
...
 
   `.start` -> `#before_consume` -> `#around_consume` [ `#consume` ] -> `.stop`
+  
+#### Batch Consumption
+
+In addition to the regular handler, Phobos provides a `BatchHandler`. The
+basic ideas are identical, except that instead of being passed a single message
+at a time, the `BatchHandler` is passed a batch of messages. All methods
+follow the same pattern as the regular handler except that they each
+end in `_batch` and are passed an array of `Phobos::BatchMessage`s instead
```
klippx: I am a bit torn in the decision to suffix with `_batch` instead of following the established Phobos standard which would be `before_consume`, `around_consume`, and `consume`. One the one hand I like that it is explicit and communicative to the user. But maybe it is enough that they have included the BatchHandler and the context they are in is already "batch" mode?
The advantage of not introducing the suffix is less complexity on the maintainer side, easier documentation, the dots feel a little bit more connected...? WDYT @jasondoc3 , maybe you want to join the discussion?

lib/phobos/actions/process_batch_inline.rb
dorner: Yep, I'm fine with that - was just mirroring the implementation in `process_message`.

README.md
klippx: Let us consider the Phobos framework from the `phobos.yml` configuration file perspective... Here, there is a mapping from topic to consumer with a configuration stating the preferred method of delivery which can be `batch`, `inline_batch` and `message`. So, for a new user of Phobos, I would recommend that if `batch` or `message` is chosen, the consumer should include only `Phobos::Handler`, and if `inline_batch` is chosen the consumer should include only `Phobos::BatchHandler`. This is not documented in the README but it is in the phobos.yml example configuration:

```
    #   - `inline_batch` will use Phobos::BatchHandler instead of Phobos::Handler to pass the entire batch to your application
    #     at once, as opposed to `batch` which will still use the standard `Phobos::Handler` to process each message
    #     individually.
```

Having this perspective it seems like an irregular use case to have a consumer that in spite of this is including both handlers, and introducing some extra configuration (not covered by Phobos framework) trying to figure out if it should use `consume` or `consume_batch`. It sounds like it is making it more complicated than necessary.

Have you investigated the possibility to put your application's consumer logic somewhere else? I would really like to hear more details on what the pain point is that the Phobos framework cannot solve for you out of the box. I have limited insight in other peoples' Phobos implementations, and any glance of it is valued.

To be completely transparent with you I am asking this because I want to have as low friction as possible in the "core" of Phobos, and if the show stopper is this special case in your app, then maybe we can solve that together 🙏🏼.

config/phobos.yml.example
```diff
@@ -107,7 +107,11 @@ listeners:
     #  - `batch` will yield batches from Ruby Kafka using `each_batch`, and commit/heartbeat at every consumed batch.
     #     Due to implementation in Ruby Kafka, it should be noted that when configuring large batch sizes in combination with taking long time to consume messages,
     #     your consumers might get kicked from the Kafka cluster for not sending a heartbeat within the expected interval.
-    #     Take this into consideration when determining your configuration.
+    #     Take this into consideration when determining your configuration. You can send heartbeats manually while
+    #     processing your messages if necessary.
+    #   - `inline_batch` will use Phobos::BatchHandler instead of Phobos::Handler to pass the entire batch to your application
```
klippx: I just realized when I wrote my latest comment that this is not really 100% accurate. It will pass to whatever consumer is configured so maybe we should phrase this something like

```
- `inline_batch` will pass the entire batch to your consumer, so make sure that this consumer includes `Phobos::BatchHandler` instead of `Phobos::Handler`
```

README.md
klippx: I found a counter argument to myself 🤔

If we went ahead and did what I argue for here, then the only difference between `Handler` and `BatchHandler` would be pluralisation of the parameters, e.g. `payload `vs `payloads`. Which could mean we might not even need or want `BatchHandler` at all. You simply include `Phobos::Handler` in either case and just make sure you know which mode of operation you are in. Which, in fact, could potentially be a bit vague/unclear as configuration change would require implementation changes. Going from `batch` to `batch_inline` would break the consumer in a not so nice way.

So if we KEEP the proposed suffix, then going from `batch` to `batch_inline` would still break the consumer, but in a much nicer way. It should suddenly fail with `undefined method :before_consume_batch` which would tell the user to start thinking in batches instead of individual messages.

A small thing like this could make this handler a bit more special, and should make configuration mistakes less common.

dorner: Yep, exactly. My use case is that I have a single "base consumer" which can be subclassed for different schemas or different purposes. The base consumer would implement both the batch and the single methods, and depending on phobos.yml, the subclasses would just work(tm) without having to know the difference.

Essentially this is the difference between Java-style method overloading:
```java
  public void doSomething(int val);
  public void doSomething(String val);
```

and the JavaScript way:
```javascript
  function doSomething(val) {
    if (typeof(val) == 'int') { ...}
    else { ... }
}
```
I find the Java way much more explicit and an easier way to split functionality.

config/phobos.yml.example
dorner: Yep, will do!

README.md
dorner: I'll update the README though!

dorner:
@klippx documentation has been updated. I think we're good to go...?

klippx:
LGTM 😇

dorner:
Woot my first merge! :D

klippx:
Feels Good Man

PR #105, opened 2019-02-04
Add parameters in constant fetcher_max_queue_size

jbustosj:
Add parameter for control fetcher max queue size acepted for ruby-kafka

### DIFF
--- a/config/phobos.yml.example
+++ b/config/phobos.yml.example
@@ -74,6 +74,8 @@ consumer:
   offset_retention_time:
   # interval between heartbeats; must be less than the session window
   heartbeat_interval: 10
+  # max number of items in the fetch queue that
+  fetcher_max_queue_size: 100
 
 backoff:
   min_ms: 1000
--- a/lib/phobos/constants.rb
+++ b/lib/phobos/constants.rb
@@ -28,8 +28,7 @@ module Phobos
       :offset_commit_interval,
       :offset_commit_threshold,
       :heartbeat_interval,
-      :offset_retention_time,
-      :fetcher_max_queue_size
+      :offset_retention_time
     ].freeze
   end
 end
--- a/lib/phobos/constants.rb
+++ b/lib/phobos/constants.rb
@@ -9,7 +9,8 @@ module Phobos
       :offset_commit_interval,
       :offset_commit_threshold,
       :heartbeat_interval,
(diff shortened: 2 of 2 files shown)


config/phobos.yml.example
```diff
@@ -74,6 +74,8 @@ consumer:
   offset_retention_time:
   # interval between heartbeats; must be less than the session window
   heartbeat_interval: 10
+  # max number of items in the fetch queue that
```
dorner: Missing some text here?

lib/phobos/constants.rb
```diff
@@ -9,7 +9,8 @@ module Constants
       :offset_commit_interval,
       :offset_commit_threshold,
       :heartbeat_interval,
-      :offset_retention_time
+      :offset_retention_time,
+      :fetcher_max_queue_size
```
dorner: There's no unit test for this - please update listener_spec.rb.

mensfeld:
`fetcher_max_queue_size: 100` - friendly advise from karafka maintainer. Set this to 10 by default. You will thank me later ;)

config/phobos.yml.example
```diff
@@ -74,6 +74,8 @@ consumer:
...
+  fetcher_max_queue_size: 100
```
klippx: @mensfeld do you have any links/references for suggesting to set this to 10? I take your word for it, but it's always great to share knowledge 👍

mensfeld: Only couple systems that blew up :D we've changed it in soon to be released 1.3 version of Karafka. In general: fetcher queue is not only used for commands but also for temp storing prebuffered data. With all the other options set to default, it may buffer couple GB of data upfront when there's a bigger offset lag. We've killed our systems couple times because of it being set to 100 (reached more than 2GB of memory usage). Having this setting so high by default can make phobos unpredictable for inexperienced users. They may assume the lib has memory leaks while it's just prebuffering a lot of data. ref: https://github.com/karafka/karafka/issues/342

klippx: Thanks @mensfeld ❤️

mensfeld: You can always count on me :)

klippx:
@jbustosj Would you like to make any updates as requested by @dorner? 🚧

PR #112, opened 2019-05-24
Allow sync producers to be cached between calls.

dorner:
In investigating the speed of our producers sending synchronously, the `shutdown` and producer creation calls turned out to have increased our Kafka connection time by around 30%. I tried this in a monkey patch and our speed increased by that amount.

This adds a new setting, `cache_sync_producer`, which makes sync producers act like async ones. It requires a `sync_producer_shutdown` call, but also caches the sync producer between calls, so it just reuses the same one and isn't constantly shutting it down and creating new ones.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,11 +55,7 @@ $ gem install phobos
 
 ### <a name="upgrade-notes"></a> Upgrade Notes
 
-Version 1.8.2 introduced a new `persistent_connections` setting for
-regular producers. This reduces the number of connections used to
-produce messages and you should consider setting it to true. This does
-require a manual shutdown call - 
-please see [Producers with persistent connections](#persistent-connection).
+Version 1.8.2 introduced a new `persistent_connections` setting for regular producers. This reduces the number of connections used to produce messages and you should consider setting it to true. This does require a manual shutdown call -  please see [Producers with persistent connections](#persistent-connection).
 
 ## <a name="usage"></a> Usage
 
--- a/README.md
+++ b/README.md
@@ -314,13 +314,15 @@ end
 
 Since the handler life cycle is managed by the Listener, it will make sure the producer is properly closed before it stops. When calling the producer outside a handler remember, you need to shutdown them manually before you close the application. Use the class method `async_producer_shutdown` to safely shutdown the producer.
 
+By default, regular producers will automatically shut themselves down after every `publish` call. You can change this behavior (which increases speed) by setting the `cache_sync_producer` config in `phobos.yml`. When set, regular producers behave identically to async producers and will also need to be shutdown manually using the `sync_producer_shutdown` method.
+
 Without configuring the Kafka client, the producers will create a new one when needed (once per thread). To disconnect from kafka call `kafka_client.close`.
 
 ```ruby
 # This method will block until everything is safely closed
 MyProducer
   .producer
-  .async_producer_shutdown
+  .async_producer_shutdown # and/or sync_producer_shutdown if regular producers are cached
 
 MyProducer
   .producer
--- a/README.md
+++ b/README.md
@@ -21,6 +21,7 @@ With Phobos by your side, all this becomes smooth sailing.
 ## Table of Contents
 
 1. [Installation](#installation)
+  1. [Upgrade Notes](#upgrade-notes)
 1. [Usage](#usage)
   1. [Standalone apps](#usage-standalone-apps)
   1. [Consuming messages from Kafka](#usage-consuming-messages-from-kafka)
@@ -51,6 +52,14 @@ Or install it yourself as:
 ```sh
 $ gem install phobos
 ```
+
+### <a name="upgrade-notes"></a> Upgrade Notes
+
+Version 1.8.2 introduced a new `persistent_connections` setting for
+regular producers. This reduces the number of connections used to
+produce messages and you should consider setting it to true. This does
+require a manual shutdown call - 
+please see [Producers with persistent connections](#persistent-connection).
 
 ## <a name="usage"></a> Usage
 
@@ -291,7 +300,7 @@ MyProducer
 
 There are two flavors of producers: __regular__ producers and __async__ producers.
 
-Regular producers will deliver the messages synchronously and disconnect, it doesn't matter if you use `publish` or `publish_list` after the messages get delivered the producer will disconnect.
+Regular producers will deliver the messages synchronously and disconnect, it doesn't matter if you use `publish` or `publish_list`; by default, after the messages get delivered the producer will disconnect.
 
 Async producers will accept your messages without blocking, use the methods `async_publish` and `async_publish_list` to use async producers.
 
@@ -310,11 +319,9 @@ class MyHandler
 end
 ```
 
-#### Note about configuring producers
+#### <a name="producer-config"></a> Note about configuring producers
 
 Since the handler life cycle is managed by the Listener, it will make sure the producer is properly closed before it stops. When calling the producer outside a handler remember, you need to shutdown them manually before you close the application. Use the class method `async_producer_shutdown` to safely shutdown the producer.
-
-By default, regular producers will automatically shut themselves down after every `publish` call. You can change this behavior (which increases speed) by setting the `cache_sync_producer` config in `phobos.yml`. When set, regular producers behave identically to async producers and will also need to be shutdown manually using the `sync_producer_shutdown` method.
 
 Without configuring the Kafka client, the producers will create a new one when needed (once per thread). To disconnect from kafka call `kafka_client.close`.
 
@@ -322,12 +329,24 @@ Without configuring the Kafka client, the producers will create a new one when n
 # This method will block until everything is safely closed
 MyProducer
   .producer
-  .async_producer_shutdown # and/or sync_producer_shutdown if regular producers are cached
+  .async_producer_shutdown
 
 MyProducer
   .producer
   .kafka_client
   .close
+```
+
+### <a name="persistent-connection"></a> Note about producers with persistent connections
+
+By default, regular producers will automatically disconnect after every `publish` call. You can change this behavior (which reduces connection overhead, TLS etc - which increases speed significantly) by setting the `persistent_connections` config in `phobos.yml`. When set, regular producers behave identically to async producers and will also need to be shutdown manually using the `sync_producer_shutdown` method.
+
+Since regular producers with persistent connections have open connections, you need to manually disconnect from Kafka when ending your producers' life cycle:
+
+```ruby
+MyProducer
+  .producer
+  .sync_producer_shutdown
 ```
 
 ### <a name="usage-as-library"></a> Phobos as a library in an existing project
--- a/config/phobos.yml.example
+++ b/config/phobos.yml.example
@@ -60,11 +60,11 @@ producer:
   # if greater than zero, the number of seconds between automatic message
   # deliveries. Only used for async_producer
   delivery_interval: 0
-  # Set this to true to cache the sync producer between publish calls.
+  # Set this to true to keep the producer connection between publish calls.
   # This can speed up subsequent messages by around 30%, but it does mean
   # that you need to manually call sync_producer_shutdown before exiting,
   # similar to async_producer_shutdown.
-  cache_sync_producer: false
+  persistent_connections: false
 
 consumer:
   # number of seconds after which, if a client hasn't contacted the Kafka cluster,
--- a/config/phobos.yml.example
+++ b/config/phobos.yml.example
@@ -60,6 +60,11 @@ producer:
   # if greater than zero, the number of seconds between automatic message
   # deliveries. Only used for async_producer
   delivery_interval: 0
+  # Set this to true to cache the sync producer between publish calls.
+  # This can speed up subsequent messages by around 30%, but it does mean
+  # that you need to manually call sync_producer_shutdown before exiting,
+  # similar to async_producer_shutdown.
+  cache_sync_producer: false
 
 consumer:
   # number of seconds after which, if a client hasn't contacted the Kafka cluster,
--- a/lib/phobos.rb
+++ b/lib/phobos.rb
@@ -55,20 +55,10 @@ module Phobos
 
     def configure(configuration)
       @config = fetch_configuration(configuration)
-      config_define_methods
+      @config.class.send(:define_method, :producer_hash) { Phobos.config.producer&.to_hash }
+      @config.class.send(:define_method, :consumer_hash) { Phobos.config.consumer&.to_hash }
       @config.listeners ||= []
       configure_logger
-    end
-
-    def config_define_methods # rubocop:disable Metrics/AbcSize
-      @config.class.send(:define_method, :cache_sync_producer) do
-        hash = Phobos.config.producer&.to_hash
-        hash[:cache_sync_producer]
-      end
-      @config.class.send(:define_method, :producer_hash) do
-        Phobos.config.producer&.to_hash&.except(:cache_sync_producer)
-      end
-      @config.class.send(:define_method, :consumer_hash) { Phobos.config.consumer&.to_hash }
     end
 
     def add_listeners(configuration)
--- a/lib/phobos.rb
+++ b/lib/phobos.rb
@@ -31,6 +31,21 @@ require 'phobos/executor'
 
 Thread.abort_on_exception = true
 
+# Monkey patch to fix this issue: https://github.com/zendesk/ruby-kafka/pull/732
+module Logging
+  # :nodoc:
+  class Logger
+    # :nodoc:
+    def formatter=(*args); end
+
+    # :nodoc:
+    def push_tags(*args); end
+
+    # :nodoc:
+    def pop_tags(*args); end
+  end
+end
+
 module Phobos
   class << self
     attr_reader :config, :logger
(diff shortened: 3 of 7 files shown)


dorner:
@klippx ^

lib/phobos.rb
```diff
@@ -31,19 +31,44 @@
...
     def configure(configuration)
       @config = fetch_configuration(configuration)
-      @config.class.send(:define_method, :producer_hash) { Phobos.config.producer&.to_hash }
-      @config.class.send(:define_method, :consumer_hash) { Phobos.config.consumer&.to_hash }
+      config_define_methods
       @config.listeners ||= []
       configure_logger
     end
 
+    def config_define_methods # rubocop:disable Metrics/AbcSize
```
klippx:
```
rubocop:disable Metrics/AbcSize
```

Is a smell and the naked eye can see that this method is very complex (which is what rubocop tries to find out for you) and should be avoided if possible.

lib/phobos.rb
```diff
@@ -31,19 +31,44 @@
...
+      @config.class.send(:define_method, :cache_sync_producer) do
```
klippx: When people who read this code and our readme read the word "cache" they won't be thinking about instance variables in ruby, they will think of "caching" stuff in general which makes this name strange. It would be similar to naming `consumer_hash` to `cache_consumer_hash` or even `instance_variable_consumer_hash`.

lib/phobos.rb
```diff
@@ -31,19 +31,44 @@
...
+        hash = Phobos.config.producer&.to_hash
+        hash[:cache_sync_producer]
+      end
+      @config.class.send(:define_method, :producer_hash) do
+        Phobos.config.producer&.to_hash&.except(:cache_sync_producer)
```
klippx: I dont like this exclusion of particular config, it makes code/config interaction complicated and hard to follow

lib/phobos/producer.rb
```diff
@@ -67,18 +67,33 @@ def kafka_client
           producer_store[:kafka_client]
         end
 
+        def create_sync_producer
```
klippx: https://github.com/phobos/phobos/diffs/1?base_sha=fcf14a6954061a01b63b22173dc58a5814f134be&commentable=true&pull_number=112&sha1=fcf14a6954061a01b63b22173dc58a5814f134be&sha2=97cac39158eec5dc0ffb7fec78e605e099278c9c&short_path=04c6e90&unchanged=expanded&utf8=%E2%9C%93#usage-producing-messages-to-kafka

> Regular producers will deliver the messages synchronously and disconnect, it doesn't matter if you use publish or publish_list after the messages get delivered the producer will disconnect.
> 
> Async producers will accept your messages without blocking, use the methods async_publish and async_publish_list to use async producers.

The change you have made kind of contradicts the description of the "regular" Producer as described by the Readme.

If you simply change from using "regular" producer to async producer, don't you achieve the exactly the same thing and reap the 30% speed increase without sacrificing code simplicity?

lib/phobos.rb
```diff
@@ -31,19 +31,46 @@
...
       @config = fetch_configuration(configuration)
-      @config.class.send(:define_method, :producer_hash) { Phobos.config.producer&.to_hash }
-      @config.class.send(:define_method, :consumer_hash) { Phobos.config.consumer&.to_hash }
+      config_define_methods
       @config.listeners ||= []
       configure_logger
     end
 
+    def config_define_methods # rubocop:disable Metrics/AbcSize
+      @config.class.send(:define_method, :cache_sync_producer) do
```
klippx: I suggest to rename this `persistent_connections` and default it to `false`. And if the connection is missing we can add some post install message that users look into this new feature and consider setting it to false (default) or migrate to true (requiring some minor changes). 

These minor changes requierd could described in a new README section:
```
1. [Installation](#installation)
  1. [Upgrade notes](#upgrading)
```

lib/phobos.rb
```diff
@@ -31,19 +31,46 @@
...
+        hash = Phobos.config.producer&.to_hash
+        hash[:cache_sync_producer]
+      end
+      @config.class.send(:define_method, :producer_hash) do
+        Phobos.config.producer&.to_hash&.except(:cache_sync_producer)
```
klippx: I don't think we need to exclude this option here, it just increases ABC size, and it is confusing that what we set in config yml doesnt show up in Phobos.config

lib/phobos.rb
```diff
@@ -31,19 +31,46 @@
...
-      @config.class.send(:define_method, :consumer_hash) { Phobos.config.consumer&.to_hash }
+      config_define_methods
       @config.listeners ||= []
       configure_logger
     end
 
+    def config_define_methods # rubocop:disable Metrics/AbcSize
+      @config.class.send(:define_method, :cache_sync_producer) do
+        hash = Phobos.config.producer&.to_hash
+        hash[:cache_sync_producer]
```
klippx: I suggest to rename this `persistent_connections` and default it to `false`. And if the connection is missing we can add some post install message that users look into this new feature and consider setting it to false (default) or migrate to true (requiring some minor changes). 

These minor changes requierd could described in a new README section:
```
1. [Installation](#installation)
  1. [Upgrade notes](#upgrading)
```

lib/phobos.rb
klippx: Would it be enough to call Phobos.config.producer&.cache_sync_producer where we need to read this, and not making this a root level config method?

config/phobos.yml.example
```diff
@@ -60,6 +60,11 @@ producer:
   # if greater than zero, the number of seconds between automatic message
   # deliveries. Only used for async_producer
   delivery_interval: 0
+  # Set this to true to cache the sync producer between publish calls.
+  # This can speed up subsequent messages by around 30%, but it does mean
+  # that you need to manually call sync_producer_shutdown before exiting,
+  # similar to async_producer_shutdown.
+  cache_sync_producer: false
```
klippx: I suggest to rename this `persistent_connections` and default it to `false`. And if the connection is missing we can add some post install message that users look into this new feature and consider setting it to false (default) or migrate to true (requiring some minor changes). 

These minor changes requierd could described in a new README section:
```
1. [Installation](#installation)
  1. [Upgrade notes](#upgrading)
```

config/phobos.yml.example
```diff
@@ -60,6 +60,11 @@ producer:
   # if greater than zero, the number of seconds between automatic message
   # deliveries. Only used for async_producer
   delivery_interval: 0
+  # Set this to true to cache the sync producer between publish calls.
```
klippx: `Set this to true to keep the producer connection between publish calls.`

README.md
```diff
@@ -314,13 +314,15 @@ end
 
 Since the handler life cycle is managed by the Listener, it will make sure the producer is properly closed before it stops. When calling the producer outside a handler remember, you need to shutdown them manually before you close the application. Use the class method `async_producer_shutdown` to safely shutdown the producer.
 
+By default, regular producers will automatically shut themselves down after every `publish` call. You can change this behavior (which increases speed) by setting the `cache_sync_producer` config in `phobos.yml`. When set, regular producers behave identically to async producers and will also need to be shutdown manually using the `sync_producer_shutdown` method.
```
klippx: `By default, regular producers will automatically disconnect after every `publish` call.`

klippx: `(which increases speed)` => `(which reduces connection overhead, TLS etc - which increases speed significantly)`

README.md
```diff
@@ -314,13 +314,15 @@ end
...
+
 Without configuring the Kafka client, the producers will create a new one when needed (once per thread). To disconnect from kafka call `kafka_client.close`.
 
 ```ruby
 # This method will block until everything is safely closed
 MyProducer
   .producer
-  .async_producer_shutdown
+  .async_producer_shutdown # and/or sync_producer_shutdown if regular producers are cached
```
klippx: Remove this comment as it describes async_producer

I think it is easier to add a separate note:

```
#### Note about producers with persistent connections

Some explanation here...

Since producers with persistent connections have open connections, you need to manually disconnect from kafka when ending your producers life cycle:

MyProducer
  .producer
  .sync_producer_shutdown
```

lib/phobos.rb
```diff
@@ -31,19 +31,46 @@
...
+
 module Phobos
   class << self
     attr_reader :config, :logger
     attr_accessor :silence_log
 
     def configure(configuration)
       @config = fetch_configuration(configuration)
-      @config.class.send(:define_method, :producer_hash) { Phobos.config.producer&.to_hash }
-      @config.class.send(:define_method, :consumer_hash) { Phobos.config.consumer&.to_hash }
```
klippx: With the suggestions below consider simply adding this lines back and getting rid of `config_define_methods` if you agree

lib/phobos.rb
dorner: Yeah... it was off by 0.3 and tbh I think the default settings for Rubocop are a bit too extreme. There's always a continuum of what people consider "complex". :) In this case I felt the function was simple enough to override the minor infraction, but given the changes I'm thinking of doing, this might be moot.

lib/phobos.rb
dorner: Yeah, I was thinking of this as "this is a producer configuration so it belongs under producer", but I realize that it does mean we have to tease them apart internally. What do you think about adding a `phobos_producer` config heading which would just contain this setting for now?

lib/phobos.rb
dorner: Unfortunately Phobos passes the entire config off to RubyKafka, which fails because it's not defined on the RubyKafka `producer` method. This was my attempt at doing it - there are at least two other ways (separating the configs out, or removing it from the hash at the time we create a new producer instead of here).

config/phobos.yml.example
dorner: Sounds great!

README.md
dorner: Sounds good.

dorner:
@klippx thanks a ton for the detailed review! I suggested a slightly different approach to the config above - let me know if you think that will work to simplify it.

lib/phobos.rb
klippx: I'd say go for the latter approach, it follows the existing pattern:

```ruby
      class PublicAPI
        NAMESPACE = :phobos_producer_store
        ASYNC_PRODUCER_PARAMS = [:max_queue_size, :delivery_threshold, :delivery_interval].freeze
        INTERNAL_PRODUCER_PARAMS = [:cache_sync_producer].freeze

        # ...

        def regular_configs
          Phobos.config.producer_hash
            .reject { |k, _| ASYNC_PRODUCER_PARAMS.include?(k) }
            .reject { |k, _| INTERNAL_PRODUCER_PARAMS.include?(k) }
        end

        def async_configs
          Phobos.config.producer_hash
            .reject { |k, _| INTERNAL_PRODUCER_PARAMS.include?(k) }
        end
```

lib/phobos.rb
klippx: Agreed, see other comment I made where I suggested name `INTERNAL_PRODUCER_PARAMS`

lib/phobos.rb
klippx: I think with the other changes/actions suggested the ABC size can be reduced so this could be removed

dorner:
@klippx made changes, please take another look!

dorner:
There's a spec failing but I'm not quite sure I understand what it's testing for - from what I can tell, the assertion is wrong. 🤔 
Also, I left the value as a hash in the config. It would simplify things a bit if we exposed it in a method though - what do you think?

README.md
```diff
@@ -52,6 +53,14 @@ Or install it yourself as:
 $ gem install phobos
 ```
 
+### <a name="upgrade-notes"></a> Upgrade Notes
+
+Version 1.8.2 introduced a new `persistent_connections` setting for
```
klippx: Nit: Consider removing manual linebreaks as the rest of the readme uses long lines + wrapping to display

lib/phobos/producer.rb
```diff
@@ -52,6 +52,7 @@ def producer
       class PublicAPI
         NAMESPACE = :phobos_producer_store
         ASYNC_PRODUCER_PARAMS = [:max_queue_size, :delivery_threshold, :delivery_interval].freeze
+        INTERNAL_PRODUCER_PARAMS = [:persistent_connections].freeze
```
klippx: I know this was my suggestion, but I was just thinking when I see the idea in code if this should be called `SYNC_PRODUCER_PARAMS` 🤔

Either one is fine.

spec/lib/phobos/producer_spec.rb
```diff
@@ -151,7 +197,11 @@ class TestProducer1
       end
 
       describe 'with a delivery threshold set' do
-        let(:config_hash) { Phobos.config.producer_hash.merge(delivery_threshold: 10) }
+        let(:config_hash) do
+          hash = Phobos.config.producer_hash.merge(delivery_threshold: 10)
+          hash.delete(:persistent_connections)
```
klippx: Is this still necessary now that we `.reject { |k, _| INTERNAL_PRODUCER_PARAMS.include?(k) }` ?

lib/phobos/producer.rb
dorner: I think I'll keep it as INTERNAL in case we want to add additional properties in the future that may or may not be related to sync producing but shouldn't be forwarded to RubyKafka.

dorner:
@klippx one more commit for you - this should fix everything!

spec/lib/phobos/producer_spec.rb
```diff
@@ -129,7 +171,11 @@ class TestProducer1
       end
 
       describe 'with a delivery interval set' do
-        let(:config_hash) { Phobos.config.producer_hash.merge(delivery_interval: 10) }
+        let(:config_hash) do
+          hash = Phobos.config.producer_hash.merge(delivery_threshold: 10)
+          hash.delete(:persistent_connections)
```
klippx: I dont think you need to do this, now that we `.reject { |k, _| INTERNAL_PRODUCER_PARAMS.include?(k) }`, right?

spec/lib/phobos/producer_spec.rb
```diff
@@ -224,9 +274,10 @@ class TestProducer1
 
     describe 'without a kafka_client configured' do
       it 'creates a new client and an async_producer bound to the current thread' do
+        config = Phobos.config.producer_hash.reject { |k, _| %i(persistent_connections).include?(k) }
```
klippx: Are you able to use the constant `INTERNAL_PRODUCER_PARAMS` here?

spec/lib/phobos/producer_spec.rb
```diff
@@ -280,6 +331,28 @@ class TestProducer1
     end
   end
 
+  describe '.sync_producer_shutdown' do
+    let(:producer) { double('Kafka::RegularProducer', produce: true, deliver_messages: true) }
+    let(:kafka_client) { double('Kafka::Client', producer: producer, close: true) }
+
+    it 'calls shutdown in the configured client and cleans up producer' do
+      original_hash = Phobos.config.producer_hash
```
klippx: consider doing this in a let

`let(:config) { Phobos.config.producer_hash.merge(:persistent_connections => true) }`

and then

`allow(Phobos.config).to receive(:producer_hash).and_return(config)`

spec/lib/phobos/producer_spec.rb
```diff
@@ -48,29 +48,71 @@ class TestProducer1
...
+          subject.producer.publish_list([
+                                          { payload: 'message-1', topic: 'topic-1', key: 'key-1', partition_key: 'part-key-1' },
+                                          { payload: 'message-2', topic: 'topic-2', key: 'key-2' }
+                                        ])
+        end
+      end
+
+      context 'with cached producer' do
+        it 'publishes and delivers a list of messages twice' do
+          original_hash = Phobos.config.producer_hash
```
klippx: similar to other comment I suggest a let

`let(:config) { Phobos.config.producer_hash.merge(persistent_connections: true) }`

klippx:
There are a few open comments, one of them old and since this was overlooked I checked other places in test suite for similar things and rounded up a new couple of comments, hope that feels good 🐸

spec/lib/phobos/producer_spec.rb
klippx: ping: old comment

spec/lib/phobos/producer_spec.rb
dorner: Nope, I still have to do this because it gets passed into the Kafka producer, which crashes on unknown keyword if I leave it in.

spec/lib/phobos/producer_spec.rb
dorner: yep, will fix

dorner:
@klippx one last round, I hope! Thanks for your patience!

klippx:
Before merging, consider if the "monkey patch"

```
Logging.init :debug, :info, :warn, :error, :fatal

# Monkey patch to fix this issue: https://github.com/zendesk/ruby-kafka/pull/732
module Logging
  # :nodoc:
  class Logger
    # :nodoc:
    def formatter=(*args); end

    # :nodoc:
    def push_tags(*args); end

    # :nodoc:
    def pop_tags(*args); end
  end
end
```

should make it into master (and released on rubygems) or if this should be opted in to by monkey patching in user land. I almost lean towards the latter but you might be more up to speed on the issue.

dorner:
@klippx in normal monkey patching I'd definitely tend towards opt-in, but in this case AFAICT there is no downside to adding these methods (they don't do anything and don't override any existing behavior), and I can't see any case where it would break anything. I'll leave the final decision up to you.

klippx:
If it solves an open issue, then yes, sure. But it would be nice to have documented confirmations that it actually solves the issue 👍

dorner:
@klippx see comment here: https://github.com/phobos/phobos/issues/111

klippx:
I merged, you can release! ✋

dorner:
Awesome, one release coming up! Thanks so much for your patience!

PR #115, opened 2019-09-16
Add support for producing and consuming messages with headers

cassiomarques:
This PR adds support for producing and consuming messages with headers.

### Producing

When producing a single message, the `publish` method signature is as follows:

```ruby
def publish(topic, payload, key = nil, partition_key = nil, headers = nil)
end
```

The difference between `key` and `partition_key` here is not clear to me and I appreciate that adding yet another positional argument to the method is not ideal, but I think changing this to use keyword arguments would be outside of the scope of this PR. 

The downside of this implementation is that if users want to specify message headers but do not care about `partition_key`, they will still have to call `publish` with all the arguments, using a value of `nil` for the `partition_key` attribute.

When publishing a batch of messages the API works better, since each message is represented by a `Hash`. I simply added one more key `headers` to the hash. 

### Consuming

~To avoid introducing breaking changes to the handler API, I decided to add a new listener configuration option to the `phobos.yml` file, called `consume_message_headers`, which defaults to `false`.~

~If `consume_message_headers` is set to to `true`, handlers need to implement the `consume` method with an additional argument for the headers:~

~If the `consume_message_headers` option is set to `false` or not specified at all, nothing changes and the version of `consume` with two arguments will be used. This guarantee compatibility for existent handlers.~

Headers will be merged into the `metadata` hash, under a `headers` key.

When consuming a batch of messages, the headers will also be present in instances of `BatchMessage`. This was actually easy to implement since in runtime `BatchMessage` simply wraps instances of `Kafka::Protocol::Record`, which will always contain the headers anyway (feature already supported by `ruby-kafka`).

I ~updated the `phobos.yml.example` file and also~ added some new stuff to the `README`.

Let me know what you think. Thanks for creating Phobos!

### DIFF
--- a/spec/lib/phobos/producer_spec.rb
+++ b/spec/lib/phobos/producer_spec.rb
@@ -19,7 +19,7 @@ RSpec.describe Phobos::Producer do
 
       expect(TestProducer1.producer)
         .to receive(:publish_list)
-        .with([{ topic: 'topic', payload: 'message', key: 'key', partition_key: nil }])
+        .with([{ topic: 'topic', payload: 'message', key: 'key', partition_key: nil, headers: nil }])
 
       subject.producer.publish('topic', 'message', 'key')
     end
@@ -32,10 +32,10 @@ RSpec.describe Phobos::Producer do
 
       expect(TestProducer1.producer)
         .to receive(:async_publish_list)
-        .with([{ topic: 'topic', payload: 'message', key: 'key', partition_key: 'partition_key' }])
+        .with([{ topic: 'topic', payload: 'message', key: 'key', partition_key: 'partition_key', headers: { foo: 'bar' } }])
 
       TestProducer1.producer.create_async_producer
-      subject.producer.async_publish('topic', 'message', 'key', 'partition_key')
+      subject.producer.async_publish('topic', 'message', 'key', 'partition_key', foo: 'bar')
     end
   end
 
@@ -58,11 +58,11 @@ RSpec.describe Phobos::Producer do
 
           expect(producer)
             .to receive(:produce)
-            .with('message-1', topic: 'topic-1', key: 'key-1', partition_key: 'part-key-1')
+            .with('message-1', topic: 'topic-1', key: 'key-1', partition_key: 'part-key-1', headers: nil)
 
           expect(producer)
             .to receive(:produce)
-            .with('message-2', topic: 'topic-2', key: 'key-2', partition_key: 'key-2')
+            .with('message-2', topic: 'topic-2', key: 'key-2', partition_key: 'key-2', headers: { foo: 'bar' })
 
           expect(producer).to receive(:deliver_messages)
           expect(producer).to receive(:shutdown)
@@ -70,7 +70,7 @@ RSpec.describe Phobos::Producer do
 
           subject.producer.publish_list([
                                           { payload: 'message-1', topic: 'topic-1', key: 'key-1', partition_key: 'part-key-1' },
-                                          { payload: 'message-2', topic: 'topic-2', key: 'key-2' }
+                                          { payload: 'message-2', topic: 'topic-2', key: 'key-2', headers: { foo: 'bar' } }
                                         ])
         end
       end
@@ -91,15 +91,15 @@ RSpec.describe Phobos::Producer do
 
           expect(producer)
             .to receive(:produce)
-            .with('message-1', topic: 'topic-1', key: 'key-1', partition_key: 'part-key-1')
+            .with('message-1', topic: 'topic-1', key: 'key-1', partition_key: 'part-key-1', headers: nil)
 
           expect(producer)
             .to receive(:produce)
-            .with('message-2', topic: 'topic-2', key: 'key-2', partition_key: 'key-2')
+            .with('message-2', topic: 'topic-2', key: 'key-2', partition_key: 'key-2', headers: { foo: 'bar' })
 
           expect(producer)
             .to receive(:produce)
-            .with('message-3', topic: 'topic-3', key: 'key-3', partition_key: 'part-key-3')
+            .with('message-3', topic: 'topic-3', key: 'key-3', partition_key: 'part-key-3', headers: nil)
 
           expect(producer).to receive(:deliver_messages).twice
           expect(producer).to_not have_received(:shutdown)
@@ -107,7 +107,7 @@ RSpec.describe Phobos::Producer do
 
           subject.producer.publish_list([
                                           { payload: 'message-1', topic: 'topic-1', key: 'key-1', partition_key: 'part-key-1' },
-                                          { payload: 'message-2', topic: 'topic-2', key: 'key-2' }
+                                          { payload: 'message-2', topic: 'topic-2', key: 'key-2', headers: { foo: 'bar' } }
                                         ])
           subject.producer.publish_list([
                                           { payload: 'message-3', topic: 'topic-3', key: 'key-3', partition_key: 'part-key-3' }
@@ -132,18 +132,18 @@ RSpec.describe Phobos::Producer do
 
         expect(producer)
           .to receive(:produce)
-          .with('message-1', topic: 'topic-1', key: 'key-1', partition_key: 'part-key-1')
+          .with('message-1', topic: 'topic-1', key: 'key-1', partition_key: 'part-key-1', headers: { foo: 'bar' })
 
         expect(producer)
           .to receive(:produce)
-          .with('message-2', topic: 'topic-2', key: 'key-2', partition_key: 'key-2')
+          .with('message-2', topic: 'topic-2', key: 'key-2', partition_key: 'key-2', headers: nil)
 
         expect(producer).to receive(:deliver_messages)
         expect(producer).to receive(:shutdown)
         expect(kafka_client).to_not receive(:close)
 
         subject.producer.publish_list([
-                                        { payload: 'message-1', topic: 'topic-1', key: 'key-1', partition_key: 'part-key-1' },
+                                        { payload: 'message-1', topic: 'topic-1', key: 'key-1', partition_key: 'part-key-1', headers: { foo: 'bar' } },
                                         { payload: 'message-2', topic: 'topic-2', key: 'key-2' }
                                       ])
       end
@@ -164,11 +164,11 @@ RSpec.describe Phobos::Producer do
 
         expect(producer)
           .to receive(:produce)
-          .with('message-1', topic: 'topic-1', key: 'key-1', partition_key: 'key-1')
+          .with('message-1', topic: 'topic-1', key: 'key-1', partition_key: 'key-1', headers: { foo: 'bar' })
 
         expect(producer)
           .to receive(:produce)
-          .with('message-2', topic: 'topic-2', key: 'key-2', partition_key: 'key-2')
+          .with('message-2', topic: 'topic-2', key: 'key-2', partition_key: 'key-2', headers: nil)
 
         expect(producer).to_not receive(:close)
       end
@@ -191,7 +191,7 @@ RSpec.describe Phobos::Producer do
             TestProducer1.producer.configure_kafka_client(kafka_client)
 
             subject.producer.async_publish_list([
-                                                  { payload: 'message-1', topic: 'topic-1', key: 'key-1' },
+                                                  { payload: 'message-1', topic: 'topic-1', key: 'key-1', headers: { foo: 'bar' } },
                                                   { payload: 'message-2', topic: 'topic-2', key: 'key-2' }
                                                 ])
           end.join
@@ -216,7 +216,7 @@ RSpec.describe Phobos::Producer do
             TestProducer1.producer.configure_kafka_client(kafka_client)
 
             subject.producer.async_publish_list([
-                                                  { payload: 'message-1', topic: 'topic-1', key: 'key-1' },
+                                                  { payload: 'message-1', topic: 'topic-1', key: 'key-1', headers: { foo: 'bar' } },
                                                   { payload: 'message-2', topic: 'topic-2', key: 'key-2' }
                                                 ])
           end.join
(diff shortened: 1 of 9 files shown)


cassiomarques:
Codeclimate doesn't like the version of `publish` with 5 arguments (I don't like it either!) but as mentioned in the PR's description, refactoring this to use keyword arguments is outside the scope of this PR. Happy to try and create another PR fixing that later!

phobos.gemspec
```diff
@@ -60,7 +60,7 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'simplecov'
   spec.add_development_dependency 'timecop'
 
-  spec.add_dependency 'activesupport', '>= 3.0.0'
+  spec.add_dependency 'activesupport', '>= 3.0.0', '< 6'
```
dorner: Why does this have to be < 6?

dorner:
@cassiomarques overall looks good - my main question would be if it would be simpler just to add the headers to the `metadata` hash rather than introducing a new argument to the consume function. I have a framework built on top of Phobos which this will significantly complicate.

I do agree that we have to change the publish method to take keyword args. I have a PR I want to add to allow the `partition` as well as `partition_key` argument and it's already too large to handle that. I can work on that or you can take a stab at it.

Btw, the difference between "key" and "partition key" is that the key is actually part of the record and e.g. is used for compaction. The partition key can be used when the record itself does not have a key, but you still want to partition your data a certain way (e.g. you have a number of events that you want to partition by user - in this case you do not want the actual key of the record to be the user ID because that implies that there is at most one record per user, which isn't true).

phobos.gemspec
cassiomarques: The latest version before the 6 series would be 5.2.3. In 6.0.0 it looks like `try` is no longer available, so things break. We could try and change all calls to `try` to use the safe navigator (but I think this is more general refactoring that's not part of the scope of this PR).

cassiomarques:
> my main question would be if it would be simpler just to add the headers to the metadata hash rather than introducing a new argument to the consume function. I have a framework built on top of Phobos which this will significantly complicate.

As far as I understand it, the metadata is not message specific, it contains information about the topic, etc. The headers are message specific and can actually change from message to message. Having those two levels of information mixed together would be confusing and would also be a departure from the way things work in `ruby-kafka`. I believe following a similar design to what `ruby-kafka` does is more intuitive. 

> I do agree that we have to change the publish method to take keyword args. I have a PR I want to add to allow the partition as well as partition_key argument and it's already too large to handle that. I can work on that or you can take a stab at it.

I'm happy to work on a new PR that changes the method signature to use keyword arguments. The main downside for this is that would introduce a breaking change, people upgrading Phobos after this change will need to update their handlers. Not sure what policy for breaking changes is being used for Phobos. 

Thanks for reviewing!

dorner:
@cassiomarques so every message has metadata for it (e.g. the key and offset) which gets passed into the handler in some way:
* For batch or message handlers, it's in the `metadata` argument.
* For inline batches, it's part of the BatchMessage.

So we should be able to add the headers per message in the same way rather than as a new argument.

dorner:
I know the keyword args will be a breaking change - we can discuss with @klippx once we've got that PR ready. I don't think we should shy away from breaking changes as long as we bump the version number up (we can decide if 1.9 or 2.0 is necessary).

cassiomarques:
@dorner When consuming a batch, won't the handler receive a list of messages (the batch) and a single metadata hash? If that's the case it wouldn't work because the headers can be different for each message. My implementation works fine for inline batches because they use `BatchMessage`, so the headers are just one more property of the object (no additional method argument problem in this case).

dorner:
@cassiomarques from what I can tell, all this information should already be present in what's passed to the consumer:
`Phobos::Actions::ProcessBatch` gets passed the batch yielded from `consumer.each_batch`
According to RubyKafka, this batch is of type `FetchedBatch`, which has an attribute `messages`
Each of these messsages is of type `FetchedMessage`, which has the `key`, `offset` and `headers` attributes.

cassiomarques:
@dorner But does Phobos call the handler's `consume` method to process each message in that batch when using `Phobos::Actions::ProcessBatch`?

dorner:
Yep, here's the code that does it: https://github.com/phobos/phobos/blob/master/lib/phobos/actions/process_batch.rb#L23
`@message` in ProcessMessage contains the headers you need. So all you need to do is modify ProcessMessage to add those headers to the metadata of the message being handled.

cassiomarques:
Right, that's actually what I'm using (but from within `Actions::ProcessMessage`) to make the headers available to the `consume` method as an additional parameter. 

I'm happy to change the implementation to use the metadata, even though I don't agree they are concepts that should be mixed up :-) I'll push some changes between and tomorrow. 

Thanks!

klippx:
First of all, thank you for trying to add headers, it is an important feature.

So let me chime in to the discussion here without looking too much at the PR itself. I feel like there is some technical debt in Phobos that surfaces more clearly to me in the light of this discussion.

1. We have too much flags to "enable" or "disable" paths in the code, these are introduced whenever we want to skip a refactor and just make our new use case work.

2. Positional parameters in `publish`, `async_publish` (but `publish_list` and `async_publish_list` are both fine). Now we are also faced with a choice to go at the handler methods by introducing more positional arguments there as well which feels like we can do better.

So to go in to more detail about the PR:
- Can we get rid of `consume_message_headers` and assume this will always be yielded in `@metadata`? The consumers implementation of handler knows best if it is safe to access `@metadata['headers']` or not. 
- At the same time can we also put headers in metadata like I believe @dorner mentioned? I think makes sense or what do you think?

```
module Phobos
  module Actions
    class ProcessMessage
      include Phobos::Processor

      attr_reader :metadata

      def initialize(listener:, message:, listener_metadata:)
        @listener = listener
        @message = message
        @metadata = listener_metadata.merge(
          key: message.key,
          partition: message.partition,
          offset: message.offset,
          headers: message.headers || {}, # ...not sure if defaulting to {} is actually needed.
          retry_count: 0
        )
      end
```

Then we solve the problem of [1] where we introduce complexity by enable/disable code paths based on config. We also solve the problem of introducing more parameters to handlers consume methods which is related to [2].

For solving the gordian knot of [2] but for the producer part I propose that we introduce new methods that use named argument and also put deprecation messages in `publish` and `async_publish` to get people to stop using it (so we can delete `publish` and `async_publish` in future versions of Phobos).

```
        def publish_message(topic: topic, payload: payload, key: key,
                            partition_key: partition_key, headers: headers)
          publish_list([{ topic: topic, payload: payload, key: key,
                          partition_key: partition_key }])
        end

        def publish(topic, payload, key = nil, partition_key = nil)
          Phobos.deprecate('publish is being replaced, please update your producers to use publish_message instead')
          publish_list([{ topic: topic, payload: payload, key: key,
                          partition_key: partition_key }])
        end
```

We should also update documentation to only talk about `publish_message`  and not mention `publish` (same for async publish).

WDYT, is this a way forward?

cassiomarques:
Hi @klippx 👋 Thanks for chiming in. I agree that merging the headers into the metadata solves the problem of having to add one more configuration option and makes Phobos easier to use. 

Something that confused me a bit when reading the code and making the changes is the fact that  `Actions::ProcessBatchInline` works based on the `ruby-kafka` message objects (which already contain the headers) but the other actions splat that object into individual values (payload, etc). I believe another option would be for handlers to always receive an instance of the `ruby-kafka` message object, regardless of the type of action. That way, all actions would have a very similar interface and receive only two arguments (the message and the topic's metadata). 

I imagine that this could cause problems to some existent code that's built on top of Phobos, so a new version would need to be released.

I'm happy to do some work to achieve what was proposed in your comment, just wondering what you and @dorner think of this other option I'm proposing here. Would it work? Would it be better? Would it break other things?

dorner:
@klippx I'm OK with deprecating the current way of accessing the methods but I don't agree with introducing new method names. We called it `publish` for a reason - calling the method `publish_messages` is a longer name for no good (permanent) reason. If we really want to deprecate current behavior, we can add both positional and keyword args to the same method:
```ruby
def def publish(p_topic=nil, p_payload=nil, p_key = nil, p_partition_key = nil, topic:nil, payload:nil, key:nil, partition_key: nil)
  raise 'Must provide a topic to the publish method!' if p_topic.nil? && topic.nil?
  if p_payload # positional args
    Phobos.deprecate('publish using positional arguments is deprecated, please use keyword arguments instead!')
  end
  publish_list([{ topic: topic || p_topic, 
                         payload: payload || p_payload, 
                         key: key || p_key,
                         partition_key: partition_key || p_partition_key }])
end  
```

But honestly I think we should start thinking about working towards a major version. Breaking changes as a way to tackle tech debt or add new features are totally fine as long as we bump the major version up as a way to indicate those changes and show people how to upgrade. There is existing deprecated functionality we want to remove as well as making changes like this - I do agree with @cassiomarques's idea of standardizing the messages being passed around as well.

What do you think?

klippx:
> But honestly I think we should start thinking about working towards a major version. Breaking changes as a way to tackle tech debt or add new features are totally fine as long as we bump the major version up as a way to indicate those changes and show people how to upgrade. There is existing deprecated functionality we want to remove as well as making changes like this - I do agree with @cassiomarques's idea of standardizing the messages being passed around as well.
> 
> What do you think?

I also think we should work toward a new major version; I would love to see the existing deprecation messages disappear with their corresponing old code.

> If we really want to deprecate current behavior, we can add both positional and keyword args to the same method

The idea of new method was to figure out how to get people to see the method is getting it's signature changed. But I agree it is better to keep the name, and if using both is a necessary evil (because god that looks nasty) to get the warning out to people then I can back it.

Historically in Phobos we have tried to be as gracious as possible and not break phobos user's code on upgrades. I like the deprecation messages because it shows good intent from developers and give users time to plan and prioritize this change. 

I think the best way forward is to go for 1.9.x with horrific method signature and then a new 2.x release which removes ALL deprecation messages including the one in 1.9.x.

> I do agree with @cassiomarques's idea of standardizing the messages being passed around as well.

I think so too, that sounds like a terrific idea @cassiomarques! It does however sound like this potentially might introduce breaking changes (?) and like a separate concern (although related). I believe that we would be better off if this refactor/proof of concept was done in a separate PR to keep changes isolated from each other and reviews/discussions more focused. WDYT? If you feel up to it 💪✨

So to summarize what it sounds like we agree on:
1. Get headers done on "existing" implementation and introduce deprecation warnings in `publish` etc in _this_ PR. This change we schedule for release 1.9.x
2. Start working on a refactor/proof of concept for how message objects are passed around, possibly also with deprecation messages if we are changing signatures. This change we schedule for release 1.9.x
3. Get rid of all deprecations and release 2.x with new method signatures from [1], [2] and all the olden golden ones that has been around for a year or so :)

What do you guys say? 🦄

cassiomarques:
@klippx That sounds like a plan :-)

Just to clarify, the scope of this PR in particular would be to cover item 1 in your list (get headers done in "existing" implementation) and items 2 and 3 would be done in subsequent PRs, is that correct?

dorner:
@klippx Sounds good to me. My only question would be - if we know we're working towards a 2.x release, shouldn't we target new features to the 2.x release instead of 1.9? This way we won't have to do the work twice for new features - just get them on the newest release. If people want to make use of them, they need to upgrade (and change whatever method signatures are necessary).

If that's the case, the order should be [2], [3], [1] in your list, no?

klippx:
Yes that was my idea 👍🏼

fre 20 sep. 2019 kl. 12:11 skrev Cássio Marques <notifications@github.com>:

klippx:
My idea was 1,2,3 to get the 1 new feature out in 1.9. Feature 2 I am not
sure if this is internal refactor Only. Maybe not a “feature”. And getting
1 done is easy since it is this pr itself 👌🏼

fre 20 sep. 2019 kl. 15:43 skrev Daniel Orner <notifications@github.com>:

dorner:
All righty, since we already have the PR. :) But I'd like to get the 2.x branch up soon-ish so we can start targeting features to it. Basically once this is merged I want to get 2 and 3 done fast. :)

cassiomarques:
Hello @dorner and @klippx 👋 

I've updated the PR, replacing the previous implementation with one that simply merges the headers into the `metadata` hash when `Actions::ProcessMessage` is used. For `Actions::ProcessBatchInline` I kept the previous implementation since we have access to headers for free there. 

I made some small updates to the README file to reflect the changes.

Codeclimate is still complaining about the `publish` methods that are now receiving too many arguments. I think we can fix that in one of the follow up PRs, making the code use keyword arguments.

Please let me know if you have any questions/comments/concerns!

Thanks!

spec/lib/phobos/producer_spec.rb
```diff
@@ -234,7 +234,7 @@ class TestProducer1
             TestProducer1.producer.configure_kafka_client(kafka_client)
 
             subject.producer.async_publish_list([
-                                                  { payload: 'message-1', topic: 'topic-1', key: 'key-1' },
+                                                  { payload: 'message-1', topic: 'topic-1', key: 'key-1', headers: { foo: 'bar' } },
```
dorner: We need to make sure we still have backwards-compatible tests where e.g. headers is not passed in.

dorner:
@cassiomarques looks good! Had one comment on the specs. For the CodeClimate issue, we can exclude that one line.

cassiomarques:
> @cassiomarques looks good! Had one comment on the specs. For the CodeClimate issue, we can exclude that one line.

According to [CodeClimate](https://codeclimate.com/github/phobos/phobos/pull/115) it's actually 4 lines 🙈

cassiomarques:
@dorner @klippx Is there anything else you need me to do before this can get merged? Let me know! Thanks!

dorner:
@klippx I'd like to merge #116 first - we can do a release with what we have and then add this to a 1.9 beta version. The PR looks good to me though.

phobos.gemspec
klippx: Resolved @dorner ?

dorner: The issue isn't with `try` (that isn't removed), it's due to certain things not being included when loading specs. See here: https://github.com/phobos/phobos/pull/117#discussion_r338227720

dorner: @cassiomarques this needs to be fixed. I think this is the last thing stopping this from being released!

cassiomarques: @dorner Sure, will work on it at some point this week and get it fixed. Thanks!

cassiomarques: @dorner I've rebased my branch and removed the commit that was limiting the `activesupport` version to `< 6`. Let me know if there are any other adjustments needed. Thanks!

dorner:
LGTM! :D

cassiomarques:
🎉

PR #117, opened 2019-10-10
Trigger heartbeat if necessary after each message in a batch.

dorner:
We have run into multiple issues where processing a batch that takes too long causes the consumer to be kicked out of the group. I know this is called out in the sample config, but I don't really see a point in us *not* heartbeating automatically after every message in batch (as opposed to inline_batch) mode. The configured heartbeat interval ensures that we're not sending heartbeats unnecessarily, and it ensures that as long as we can get through individual messages fast enough, we should be good forever.

@klippx we can aim for 1.9 with this, no?

### DIFF
--- a/config/phobos.yml.example
+++ b/config/phobos.yml.example
@@ -109,11 +109,8 @@ listeners:
     # Possible values: [`message`, `batch` (default)]
     #  - `message` will yield individual messages from Ruby Kafka using `each_message` and will commit/heartbeat at every consumed message.
     #     This is overall a bit slower than using batch, but easier to configure.
-    #  - `batch` will yield batches from Ruby Kafka using `each_batch`, and commit/heartbeat at every consumed batch.
-    #     Due to implementation in Ruby Kafka, it should be noted that when configuring large batch sizes in combination with taking long time to consume messages,
-    #     your consumers might get kicked from the Kafka cluster for not sending a heartbeat within the expected interval.
-    #     Take this into consideration when determining your configuration. You can send heartbeats manually while
-    #     processing your messages if necessary.
+    #  - `batch` will yield batches from Ruby Kafka using `each_batch`, and commit at every consumed batch. It will
+    #     still heartbeat after every message if necessary (using the heartbeat_interval, below).
     #   - `inline_batch` also uses `each_batch`, but will pass the entire batch to your handler instead
     #     of one message at a time. To use this method, you should include Phobos::BatchHandler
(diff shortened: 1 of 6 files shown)


config/phobos.yml.example
```diff
@@ -109,11 +109,8 @@ listeners:
     # Possible values: [`message`, `batch` (default)]
     #  - `message` will yield individual messages from Ruby Kafka using `each_message` and will commit/heartbeat at every consumed message.
     #     This is overall a bit slower than using batch, but easier to configure.
-    #  - `batch` will yield batches from Ruby Kafka using `each_batch`, and commit/heartbeat at every consumed batch.
-    #     Due to implementation in Ruby Kafka, it should be noted that when configuring large batch sizes in combination with taking long time to consume messages,
-    #     your consumers might get kicked from the Kafka cluster for not sending a heartbeat within the expected interval.
-    #     Take this into consideration when determining your configuration. You can send heartbeats manually while
-    #     processing your messages if necessary.
+    #  - `batch` will yield batches from Ruby Kafka using `each_batch`, and commit at every consumed batch. It will
+    #     still heartbeat after every message if necessary (using the heartbeat_interval, below).
```
klippx: I don't see a problem changing this either, but just out of curiosity - for your particular problem did you analyse if you indeed have too large batch sizes or not? I didn't see any references to this in your analysis. In general it feels like good recommendation to decrease batch size with respect to how long processing each individual message takes.

I can recommend https://github.com/phobos/phobos_prometheus or something similar to get a good picture on how your consumption is doing.

spec/spec_helper.rb
```diff
@@ -18,6 +18,8 @@
 require 'pry-byebug'
 require 'timecop'
 require 'phobos/test'
+require 'active_support/core_ext'
+require 'active_support/json'
```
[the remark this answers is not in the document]
dorner: I believe this was causing test failures with ActiveSupport 6 when it wasn't there.

config/phobos.yml.example
dorner: @klippx Yep, decreasing batch size usually helps, but we've run into issues when we have a sudden spike in unexpected messages on a previously low-traffic topic. This is essentially a catch-all to make sure that batch size would only affect the network traffic for fetches and committing offsets and wouldn't risk losing the session entirely. This can end up being a fatal error as the consumer keeps consuming the same too-large batch over and over again.

PR #119, opened 2019-12-09
Use keyword arguments when publshing messages

cassiomarques:
Hello 👋 

This PR changes the signature of the `publish` and `publish_async` methods to use keyword arguments instead of positional arguments.

Currently those methods accept 5 arguments, which makes it harder to know in which order those arguments should be specified. The use of keyword arguments makes it easier to use Phobos when producing messages.

### DIFF
--- a/lib/phobos.rb
+++ b/lib/phobos.rb
@@ -78,7 +78,8 @@ module Phobos
     end
 
     def deprecate(message)
-      warn "DEPRECATION WARNING: #{message} #{Kernel.caller.first}"
+      location = caller.find { |line| line !~ %r{/phobos/} }
+      warn "DEPRECATION WARNING: #{message}: #{location}"
     end
 
     # :nodoc:
--- a/lib/phobos/producer.rb
+++ b/lib/phobos/producer.rb
@@ -11,16 +11,26 @@ module Phobos
     end
 
     class PublicAPI
+      MissingRequiredArgumentsError = Class.new(StandardError) do
+        def initialize
+          super('You need to provide a topic name and a payload')
+        end
+      end
+
       def initialize(host_obj)
         @host_obj = host_obj
       end
 
-      def publish(topic, payload, key = nil, partition_key = nil, headers = nil)
-        class_producer.publish(topic, payload, key, partition_key, headers)
+      def publish(*args, **kwargs)
+        args = normalize_arguments(*args, **kwargs)
+
+        class_producer.publish(**args)
       end
 
-      def async_publish(topic, payload, key = nil, partition_key = nil, headers = nil)
-        class_producer.async_publish(topic, payload, key, partition_key, headers)
+      def async_publish(*args, **kwargs)
+        args = normalize_arguments(*args, **kwargs)
+
+        class_producer.async_publish(**args)
       end
 
       # @param messages [Array(Hash(:topic, :payload, :key, :headers))]
@@ -42,6 +52,30 @@ module Phobos
       def class_producer
         @host_obj.class.producer
       end
+
+      # rubocop:disable Metrics/ParameterLists
+      def normalize_arguments(p_topic = nil, p_payload = nil, p_key = nil,
+                              p_partition_key = nil, p_headers = {},
+                              **kwargs)
+        {}.tap do |args|
+          {
+            topic: p_topic,
+            payload: p_payload,
+            key: p_key,
+            partition_key: p_partition_key,
+            headers: p_headers
+          }.each { |k, v| args[k] = kwargs[k] || v }
+
+          raise MissingRequiredArgumentsError if [:topic, :payload].any? { |k| args[k].nil? }
+
+          kwargs.each do |k, v|
+            next if args.key?(k)
+
+            args[:headers][k] = v
+          end
+        end
+      end
+      # rubocop:enable Metrics/ParameterLists
     end
 
     module ClassMethods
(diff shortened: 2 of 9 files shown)


cassiomarques:
Hi @dorner 👋 I gave it a try and came up with something like this:

```ruby
      def publish(p_topic = nil, p_payload = nil, p_key = nil,
                  p_partition_key = nil, p_headers = nil,
                  topic: nil, payload: nil, key: nil, partition_key: nil, headers: nil)
        topic         ||= p_topic
        payload       ||= p_payload
        key           ||= p_key
        partition_key ||= p_partition_key
        headers       ||= p_headers

        raise MissingRequiredArgumentsError if [topic, payload, key].any?(&:nil?)

        class_producer.publish(
          topic: topic, payload: payload, key: key, partition_key: partition_key, headers: headers
        )
      end
```

I had to set all positional arguments as optional as well, otherwise there's no way to call the method using only keyword arguments, as the arity would be wrong.

The problem is when we want to call `publish` like this:

```ruby
subject.producer.async_publish('topic', 'message', 'key', nil, { foo: 'bar' })
```

The last argument are the headers, however Ruby has no way of telling whether that should be the `headers` argument or the start of the keyword arguments, so it complains with:

```
 ArgumentError:
       unknown keyword: foo
```

I'm not sure how to have an implementation where positional and keyword arguments coexist without having problems like the above. Any ideas?

dorner:
Funnily enough this is exactly the situation that the changes in Ruby 2.7 are meant to address :)
We can try this maybe?
```ruby
      def publish(*args, **kwargs)
        p_topic, p_payload, p_key, p_partition_key, p_headers = *args
        topic = kwargs[:topic] || p_topic
        payload = kwargs[:payload] || p_payload
        key = kwargs[:key] || p_key
        partition_key = kwargs[:partition_key] || p_partition_key
        headers = kwargs[:headers] || p_headers || {}

        raise MissingRequiredArgumentsError if [topic, payload, key].any?(&:nil?)
        kwargs.each do |k, v|
         next if %i(topic payload key partition_key headers).include?(k)
         headers[k] = v
        end

        class_producer.publish(
          topic: topic, payload: payload, key: key, partition_key: partition_key, headers: headers
        )
      end
```

cassiomarques:
@dorner But then the new version would only work with Ruby > 2.7, it would defeat the point of keeping both APIs available. Thoughts?

dorner:
@cassiomarques ah, sorry if I was unclear. This works now with Ruby 2.0. 2.7 will be making some changes but what I pasted isn't using those changes.

dorner:
Btw one thing I realized is although we would need the `**kwargs` we would not need the `*args` - you can leave them as individual optional arguments.

cassiomarques:
@dorner Thanks for your suggestion 👍 I think I've found a way to make it work ✅  Let me know what you think!

lib/phobos/producer.rb
```diff
@@ -42,6 +52,30 @@ def async_publish_list(messages)
...
+            partition_key: p_partition_key,
+            headers: p_headers
+          }.each { |k, v| args[k] = kwargs[k] || v }
+
+          raise MissingRequiredArgumentsError if [:topic, :payload].any? { |k| args[k].nil? }
+
+          kwargs.each do |k, v|
+            next if args.key?(k)
+
+            args[:headers][k] = v
```
dorner: We need to also print a deprecation message if there are no keyword arguments.

lib/phobos/producer.rb
```diff
@@ -42,6 +52,30 @@ def async_publish_list(messages)
...
+                              p_partition_key = nil, p_headers = {},
+                              **kwargs)
+        {}.tap do |args|
+          {
+            topic: p_topic,
+            payload: p_payload,
+            key: p_key,
+            partition_key: p_partition_key,
+            headers: p_headers
+          }.each { |k, v| args[k] = kwargs[k] || v }
```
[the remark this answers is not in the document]
cassiomarques: I had to do that to avoid too many explicit conditional assignments, Rubocop wasn't impressed 🤷‍♂

lib/phobos/producer.rb
cassiomarques: @dorner Done ✅ 

I had to change the implementation of `Phobos.deprecate`. It was using `Kernel.caller.first` directly, However because code inside the gem was calling that method, the caller was not the gem's client code but actually the gem itself! So the locations being printed out in the deprecation messages were misleading.

lib/phobos.rb
```diff
@@ -77,8 +77,8 @@ def create_exponential_backoff(backoff_config = nil)
       ExponentialBackoff.new(min, max).tap { |backoff| backoff.randomize_factor = rand }
     end
 
-    def deprecate(message)
-      warn "DEPRECATION WARNING: #{message} #{Kernel.caller.first}"
+    def deprecate(message, location)
+      warn "DEPRECATION WARNING: #{message}: #{location}"
```
dorner: If you're trying to avoid printing out Phobos methods, this is probably a better generalized solution (note you can omit the "Kernel" part of the call):
`caller.find { |s| s !~ /\/phobos\// }`
See if that works, in which case you don't need to change all the other calling code.

cassiomarques: Yep, this does work 👍 Updated the code accordingly!

dorner:
LGTM!

dorner:
This is going to go into a 1.9.0-beta1 as per our earlier discussion.

klippx:
Nice job!

PR #121, opened 2020-01-22
Version 2.0 - Remove deprecated behavior

dorner:
@klippx @cassiomarques here it is! Let me know what you think!

I'll create a 1.9.x branch off master before merging this so we can support existing versions long-term.

To clarify, this removes the following patterns:
* `before_consume` method
* `around_consume` as a class method or without yielding values
* `publish` and `async_publish` with positional arguments

We have instead standardized on `around_consume` as an instance method that then yields the values to the `consume` method - so we don't force mutations and have both methods be instance methods.

We also moved to keyword arguments for publishing since the publish methods now have more than the recommended number of arguments and are becoming difficult to use.

### DIFF
--- a/lib/phobos/producer.rb
+++ b/lib/phobos/producer.rb
@@ -21,18 +21,20 @@ module Phobos
         @host_obj = host_obj
       end
 
-      def publish(*args, **kwargs)
-        Phobos.deprecate(deprecate_positional_args_message('publish')) if kwargs.empty?
-
-        args = normalize_arguments(*args, **kwargs)
-        class_producer.publish(**args)
+      def publish(topic:, payload:, key: nil, partition_key: nil, headers: nil)
+        class_producer.publish(topic: topic,
+                               payload: payload,
+                               key: key,
+                               partition_key: partition_key,
+                               headers: headers)
       end
 
-      def async_publish(*args, **kwargs)
-        Phobos.deprecate(deprecate_positional_args_message('async_publish')) if kwargs.empty?
-
-        args = normalize_arguments(*args, **kwargs)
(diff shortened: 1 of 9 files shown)


lib/phobos/producer.rb
```diff
@@ -21,18 +21,20 @@ def initialize(host_obj)
         @host_obj = host_obj
       end
 
-      def publish(*args, **kwargs)
-        Phobos.deprecate(deprecate_positional_args_message('publish')) if kwargs.empty?
-
-        args = normalize_arguments(*args, **kwargs)
-        class_producer.publish(**args)
+      def publish(topic:, payload:, key: nil, partition_key: nil, headers: nil)
```
cassiomarques: We can also remove the `MissingRequiredArgumentsError` exception class 🔥

dorner: Done!

dorner:
@klippx still waiting on this - are we good to merge? If I don't hear back in a couple of days I'll complete this step.

klippx:
Do we need any README updates to reflect the changes? Also a description of the PR would be nice, it is a bit vague to say "remove stuff" when we look at the PR 1 year later.

For me who is only "theoretically" involved, would be nice to see some examples of how to migrate from common old patterns to the new preferred patterns.

A lot of old smelly code was removed, it looks good! I approve in good faith!

dorner:
@klippx thanks for the feedback - I'll add a migration section to the README and then we're good to go. 🎉

dorner:
WOOHOO!

cassiomarques:
🚀

PR #136, opened 2021-05-14
DRAFT: GUILD-625: Changes required to support multiple topic subscrip…

eduardopoleoflipp:
…tions per listener

### DIFF
--- a/lib/phobos/listener.rb
+++ b/lib/phobos/listener.rb
@@ -9,7 +9,7 @@ module Phobos
     DEFAULT_MAX_BYTES_PER_PARTITION = 1_048_576 # 1 MB
     DELIVERY_OPTS = %w[batch message inline_batch].freeze
 
-    attr_reader :group_id, :topic, :id
+    attr_reader :group_id, :topics, :id
     attr_reader :handler_class, :encoding, :consumer
 
     # rubocop:disable Metrics/MethodLength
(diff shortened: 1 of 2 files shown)


lib/phobos/listener.rb
```diff
@@ -9,7 +9,7 @@ class Listener
     DEFAULT_MAX_BYTES_PER_PARTITION = 1_048_576 # 1 MB
     DELIVERY_OPTS = %w[batch message inline_batch].freeze
 
-    attr_reader :group_id, :topic, :id
+    attr_reader :group_id, :topics, :id
```
dorner, line 12: For backwards compatibility, let's keep both `:topic` and `:topics` and deprecate `:topic`. `attr_reader` is public so we have no idea if someone would be accessing this for some reason.

lib/phobos/listener.rb
```diff
@@ -87,13 +87,17 @@ def send_heartbeat_if_necessary
...
-      { listener_id: id, group_id: group_id, topic: topic, handler: handler_class.to_s }
+      { listener_id: id, group_id: group_id, topic: topics, handler: handler_class.to_s }
     end
 
     def start_listener
       instrument('listener.start', listener_metadata) do
         @consumer = create_kafka_consumer
-        @consumer.subscribe(topic, **@subscribe_opts)
+
+        # ruby-kafka does not provide an api to subscribe a group of particular topics
```
dorner, line 97: There technically is one, although it would mean turning `topic` into a regex that includes all of the topics, e.g.
```ruby
topic_regex = Regexp.new("^#{topics.join('|')}$")
@consumer.subscribe(topic_regex, **@subscribe_opts)
```

...but I don't think that's any cleaner at all. :p Maybe just take the comment out.

dorner (changes requested):
Added some comments. Also, are there specs we can improve to cover this case?

PR #138, opened 2021-05-26
Add ability to override kafka connection config for producers or consumers separately

sesquipedalian-dev:
We have a use case where we can have usually a single consumer and multiple producers in a single process, and we're looking to migrate to a new kafka broker with different configuration settings. To support this we'd like to be able to specify different connection parameters (e.g. seed_brokers, authentication parameters) for consumer or producer, within the same config file. 

This PR has the producer and consumer add a config key to their call to `create_kafka_client` to accomplish that.  We also had to filter out the `kafka` key when creating the `consumer_hash` so that `ruby-kafka` didn't complain about the extra param.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ config/*.yml
 log/*.log
 .byebug_history
 .idea
+*.gem
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,0 +1,8 @@
+release:
+  stage: deploy
+  only:
+    - /^v[0-9.].*$/
+  script:
+    - gem build phobos.gemspec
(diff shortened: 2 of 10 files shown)


dorner:
@sesquipedalian-dev this looks great! I'd just ask you to remove the version bump since that's handled separately from the PR, and put the changelog entry under UNRELEASED. When we do the version bump we'll move it from that section to the new version.

Thanks!

sesquipedalian-dev:
@dorner thank you for looking! I've un-bumped the version.  I also tweaked the readme section for local development a little bit to use the docker-compose `test` as travis-CI does; a colleague and I both had trouble running the specs following the previous instructions.  Let me know any more tweaks!

dorner (approved):

dorner:
Looks good - thanks!

dorner:
Version bumped to 2.1.0 🎉

sesquipedalian-dev:
Thanks for the quick merge & release!

PR #141, opened 2021-09-02
Bump ruby-kafka version to 1.4 to enable different topic subscription…

eduardopoleoflipp:
… within the same consumer group


dorner:
Per discussion, this shouldn't be necessary.

PR #144, opened 2021-12-17
[ISSUE 126] recreate listener after too many retries

reachlin:
try to create a new kafka client after many retry failures,
https://github.com/phobos/phobos/issues/126

### DIFF
--- a/lib/phobos/listener.rb
+++ b/lib/phobos/listener.rb
@@ -9,7 +9,7 @@ module Phobos
     DEFAULT_MAX_BYTES_PER_PARTITION = 1_048_576 # 1 MB
     DELIVERY_OPTS = %w[batch message inline_batch].freeze
 
-    attr_reader :group_id, :topic, :id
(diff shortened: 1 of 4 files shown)


lib/phobos/listener.rb
```diff
@@ -18,13 +18,14 @@ def initialize(handler:, group_id:, topic:, min_bytes: nil, max_wait_time: nil,
...
                    heartbeat_interval: nil, offset_commit_threshold: nil,
-                   offset_retention_time: nil)
+                   offset_retention_time: nil, max_retries: -1)
       @id = SecureRandom.hex[0...6]
       @handler_class = handler
       @group_id = group_id
       @topic = topic
       @backoff = backoff
       @delivery = delivery.to_s
+      @max_retries = max_retries.to_i
```
yin-ym, line 28: if someone specified a wrong number in max_retries like string, this max_retries will be set to 0. And the retry logic above will not run as expected (but just do more retry with creating client). Seems harmless, but may need to consider.

PR #147, opened 2021-12-29
CI workflow on github actions

reachlin:
infra team recommends github actions for ci workflow, so add it


dorner:
@reachlin what's going on here? This is the third PR that was opened and then closed within a day.

dorner:
It seems like you're trying to fork phobos into a private gem? Is there a reason that's happening rather than trying to contribute back to the original project?

reachlin:
sorry, my mistake

PR #148, opened 2022-01-18
Allow actual YAML/PSYCH versions to be supported

trialize:
This will allow Psych::VERSION > '3.1.0.pre1' to be supported (is used by YAML.safe_load).

( See https://ruby-doc.org/stdlib-3.1.0/libdoc/psych/rdoc/index.html - Ruby 3.1.0 )

### DIFF
--- a/lib/phobos.rb
+++ b/lib/phobos.rb
@@ -114,14 +114,25 @@ module Phobos
     def read_configuration(configuration)
       return configuration.to_h if configuration.respond_to?(:to_h)
 
-      YAML.safe_load(
-        ERB.new(
-          File.read(File.expand_path(configuration))
-        ).result,
-        [Symbol],
-        [],
-        true
-      )
+      if Gem::Version.new(Psych::VERSION) >= Gem::Version.new('3.1.0.pre1')
+        YAML.safe_load(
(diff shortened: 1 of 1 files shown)


dorner:
@trialize can you split this method up so CodeClimate doesn't complain? :)

trialize:
Rubocop will complain anyways.

module Phobos already 100 lines long so nothing new can be added without sacrificing beauty.

dorner:
The point of having something like CodeClimate is to make sure that things don't get too complex. Right now the module is exactly 100 lines so we can't add anything more to it without doing some refactoring. For example, we could move the configuration pieces going into its own module rather than the top-level Phobos module.

trialize:
Sure, CodeClimate is definitely helpful. And sure, you could move anything anywhere for any sake you'd feel appropriate.
The intention was to point to Ruby 3.1.0 incompatibility and propose quick solution. 
Meanwhile I'll monkey-patch. Thanks!

dorner:
I've fixed this here: https://github.com/phobos/phobos/commit/1e7a081ace84c96e42d8da9920c4d2f543b96e8b

PR #154, opened 2022-05-11
allow regexp tags in config file

lukewestby:
As described in https://github.com/phobos/phobos/issues/153

I want to be forthcoming that Ruby is not one of my strong languages, so in particular if there are any ruby-specific security considerations that should go into this change please keep in mind that I may have missed them. 

Thanks for your time and effort!

### DIFF
--- a/config/phobos.yml.example
+++ b/config/phobos.yml.example
@@ -94,6 +94,8 @@ backoff:
 
 listeners:
   - handler: Phobos::EchoHandler
+    # Name of the topic. You can also specify a Regexp to listen to mulitple topics
+    # topic: !ruby/regexp /^some-prefix\./
     topic: test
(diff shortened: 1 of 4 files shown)


sbeckeriv (approved):
secure enough because we control the yaml.

config/phobos.yml.example
```diff
@@ -94,6 +94,8 @@ backoff:
 
 listeners:
   - handler: Phobos::EchoHandler
+    # Name of the topic. You can also specify a Regexp to listen to mulitple topics
```
dorner, line 97: "multiple"

dorner (changes requested):
Looks good to me! One documentation typo. Can you also add a section in the CHANGELOG, under UNRELEASED?

Issue #152, opened 2022-04-12
Phobos batch delivery kills consumer thread due to HeartBeat error


harsha-flipp:
https://github.com/phobos/phobos/blob/master/lib/phobos/actions/process_batch.rb#L29

This line could raise a HeartBeatError under certain conditions of kafka, say, [rebalanceInProgress](https://github.com/zendesk/ruby-kafka/blob/master/lib/kafka/consumer_group.rb#L105-L131).

This [causes consumer to throw ProcessingError ](https://github.com/zendesk/ruby-kafka/blob/master/lib/kafka/consumer.rb#L336,L345)which, in turn, kills the consumer_loop entirely

Expected: Kafka topic rebalancing should not raise ProcessingError.
Actual: It raises ProcessingError and end up leaving the consumer group [entirely](https://github.com/zendesk/ruby-kafka/blob/master/lib/kafka/consumer.rb#L409).

Workaround: Use batch_inline as that flow does not contain this heartbeat trigger in the action.

PR #155, opened 2022-05-13
process_batch action catches runtime error on heartbeat

harsha-flipp:
#152 Heartbeat error is manifested into Processing error even though it's not a ProcessingError.
This catches heartbeat error and logs a warning message.

I think phobos cannot take any action based on this heartbeat response. It happens at consumer level in ruby-kafka so it's best to log as a warning instead of an error indicating that something happened.


dorner (approved):

dorner:
LGTM!

PR #157, opened 2022-06-15
Raise exception so that executor restarts the thread

dorner:
We are seeing issues where a crash on business logic within a consumer causes that listener thread to crash and never come back. This seems to be because the error gets propagated all the way down to RubyKafka, which raises a ProcessingError, which is then *caught* by Phobos and swallowed. The listener is shut down and doesn't come back.

This PR re-raises the exception so that the executor can call `handle_crashed_listener`.

### DIFF
--- a/lib/phobos/listener.rb
+++ b/lib/phobos/listener.rb
@@ -42,12 +42,14 @@ module Phobos
 
     def start
       @signal_to_stop = false
+      error = nil
 
       start_listener
 
       begin
         start_consumer_loop
-      rescue Kafka::ProcessingError, Phobos::AbortError
+      rescue Kafka::ProcessingError, Phobos::AbortError => e
+        error = e
         # Abort is an exception to prevent the consumer from committing the offset.
         # Since "listener" had a message being retried while "stop" was called
         # it's wise to not commit the batch offset to avoid data loss. This will
@@ -58,6 +60,7 @@ module Phobos
       end
     ensure
       stop_listener
+      raise error if error
     end
 
     def stop
--- a/lib/phobos/listener.rb
+++ b/lib/phobos/listener.rb
@@ -42,14 +42,11 @@ module Phobos
 
     def start
       @signal_to_stop = false
-      error = nil
-
       start_listener
 
       begin
         start_consumer_loop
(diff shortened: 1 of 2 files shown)


lib/phobos/listener.rb
```diff
@@ -42,12 +42,14 @@ def initialize(handler:, group_id:, topic:, min_bytes: nil, max_wait_time: nil,
...
       @signal_to_stop = false
+      error = nil
 
       start_listener
 
       begin
         start_consumer_loop
-      rescue Kafka::ProcessingError, Phobos::AbortError
+      rescue Kafka::ProcessingError, Phobos::AbortError => e
+        error = e if e.is_a?(Kafka::ProcessingError)
```
harsha-flipp, line 52: `raise e if e.is_a?(Kafka::ProcessingError)`

```ruby
begin
   puts "begin block"
   raise 'blue'
rescue Exception
   puts "error block";
   raise 'green'
ensure
  puts "always block";
end
```
You can raise error in rescue block. Ensure block runs always.
You can avoid the error variable entirely. Just wondering if I am missing something.

harsha-flipp: I just realised the structure is little different after taking a look at the listener.rb file. However, raising an error in rescue works as expected.

```ruby
def method1
  begin
    puts 'begin block'
    raise 'BeginError'
  rescue Exception => e
    puts 'exception block'
    raise e
  end
ensure
  puts 'method ensure'
end
```
Output:
> begin block
> exception block
> method ensure
> Traceback (most recent call last):
>         5: from /Users/harsha.subraveti/.rvm/rubies/ruby-3.0.0/bin/irb:23:in `<main'`
>         4: from /Users/harsha.subraveti/.rvm/rubies/ruby-3.0.0/bin/irb:23:in `load'`
>         3: from /Users/harsha.subraveti/.rvm/rubies/ruby-3.0.0/lib/ruby/gems/3.0.0/gems/irb-1.3.0/exe/irb:11:in `<top (required)'
>         2: from (irb):12:in `<main'
>         1: from (irb):4:in `method1'
> RuntimeError (BeginError)

harsha-flipp (approved):

PR #158, opened 2022-07-27
Add RBS definitions via sord

dorner:
Add YARD comments so we can run `sord` on the codebase.

This is using my fork of `sord` that has the `--hide-private` feature: https://github.com/AaronC81/sord/pull/146


dorner:
This is part of the CI issues we're having - Travis keeps segfaulting. If you click into it you'll see that it passes on one version of Ruby but fails on the other two. Angad is looking into this part.

lib/phobos/listener.rb
```diff
@@ -9,9 +9,28 @@ class Listener
...
+    attr_reader :encoding, :consumer
+
+    # @param handler [Class]
+    # @param group_id [String]
+    # @param topic [String]
+    # @param min_bytes [Integer]
+    # @param max_wait_time [Integer]
+    # @param start_from_beginning [Boolean]
+    # @param delivery [String]
+    # @param max_bytes_per_partition [Integer]
```
jenchanflipp, line 28: Just curious - are all these annotations automatically generated? How come some of the parameters are missing? e.g. force_encoding, backoff

dorner: Nope, they're done by hand, and that's why some are missing. :) Most of the missing info comes from RubyKafka. I didn't actually want to document every type here - I'm more interested in the method definitions themselves (i.e. X object responds to Y method that takes Z parameters) so that I can use them to document Deimos. Once that's done I could move back to this and fill in whatever's missing.

jenchanflipp (approved):

PR #161, opened 2022-10-25
Merge upstream 10-25-2022

reachlin:
merge the community changes for security reasons.


reachlin:
pls ignore, we have to add a second producer config for phobos. always pr to the wrong remote :P

PR #165, opened 2023-08-31
Generate Sorbet RBI files

dorner:


dorner:
I asked this on the Sorbet slack and apparently this is the recommended approach. I added `.gitattributes` so any further changes wouldn't result in huge PRs.

